PR #16, opened 2012-03-09
Getting a PollingLabel to show itself.

m-gibson:
Hi again,

When attempting to add a PollingLabel widget to my bar, I found it wouldn't display.  XMonadLog, however, had been working fine.  I saw that they in fact both use gtk Labels, but `xmonadLogNew` runs the action `widgetShowAll l` before the `toWidget` call, while `pollingLabelNew` doesn't.  I tried adding that same line to the PollingLabel code, and that did cause the relevant widget to show up properly in my bar.  This was complete trial-and-error, but hopefully it solves a problem that could show up on others' systems.  If you want info about my configuration let me know.

Btw, thanks for merging my other request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ There are also several more specialized widgets:
  * Battery widget
  * Textual clock
  * Freedesktop.org notifications (via dbus)
- * MPRIS widget (currently only supports MPRIS1)
+ * MPRIS1 and MPRIS2 widgets
  * Weather widget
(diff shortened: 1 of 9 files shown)


mellon85:
I am having the same issue, but I solved by just setting the widget itself as visible.

```haskell
wrapLabel :: IO Widget -> IO Widget                                            
wrapLabel generator = do                                                       
    widget <- generator                                                   
    widgetShow widget                                                          
    return widget  
```

PR #64, opened 2014-01-30
Added a new battery format for showing charging status

abdulsattar:
It shows "On" for charging and shows the percentage otherwise. Made it the
default.

### DIFF
--- a/src/System/Taffybar/Battery.hs
+++ b/src/System/Taffybar/Battery.hs
@@ -33,6 +33,13 @@ battInfo ctxt fmt = do
             minutes' = minutes `mod` 60
         in printf "%02d:%02d" hours minutes'
 
+      batteryOn :: Bool
+      batteryOn = case (batteryState info) of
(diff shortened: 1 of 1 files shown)


silasdavis:
Would be really nice to see this, any comments? Very useful to see if your laptop is actually plugged in or not.

abdulsattar:
The problem with this is, it uses polling to know the status. It'd be great if we could handle an event to know the status. I don't really know how to do that.

travitch:
One possibility would be to listen for the 'changed' signal in the UPower DBus API.  Whenever the state changes, you could read the 'OnBattery' property.  The interplay with suspend/resume might be annoying, but UPower claims to send those signals, too.

http://upower.freedesktop.org/docs/UPower.html

sakshamsharma:
This PR has been superseded by #147 , although we can probably borrow ideas from this into the code, now that we have a handler for state.

PR #66, opened 2014-01-31
Workspace images

teleshoes:
1) Generalizes WSS markup functions to IO actions
2) add Image widget to Workspace buttons, and include the hbox for more general control
3) add imageSelector to allow endusers to write their own image selection functions
taffybar

this is a breaking change and it doesnt provide an elegant ootb solution for images. however, folks who dont want images dont have to do anything, and folks who do can write an imageSelector. if someone feels ambitious, a default imageSelector could be written to give a decent ootb experience.

i dont expect you to merge this, but it would be really great if you did, or impld something like it, not least because i could stop maintaining a fork.


seniorivn:
@travitch will you merge it? Or make the alternative?
@teleshoes are you still use your fork?

teleshoes:
yep, i use it every day, and i would SO DEARLY LOVE not to have to keep managing these awful merge conflicts every few months. naturally, they get worse each time.

personally, i agree that the feature is not polished enough for wide use, but i DO think that my code could form the foundation of a full-fledged impl of this feature, and its not like it hurts things. (afaik, only a handful of humans on the face of the earth use my fork)

teleshoes:
see #167

teleshoes:
#167 was merged!

PR #99, opened 2014-10-24
Rework pager formatting

bgamari:
This patch allows the user to mark up workspace names with access to
monadic context.  This enables things like,

``` haskell
    myMarkupWorkspaces :: Traversable f => f WorkspaceInfo -> f Markup
    myMarkupWorkspaces wss = evalState (traverse f wss) ("", cycle colors)
      where
        colors = [ "#843c3c"
                 , "#60843c"
                 , "#3c8384"
                 , "#3c4d84"
                 , "#613c84"
                 , "#843c3c"
                 ]

        f :: WorkspaceInfo -> State (String, [String]) Markup
        f ws@(WSInfo {wsiName=name}) = do
          let group = takeWhile (/= '-') name
          (lastGroup, grpColor:rest) <- get
          if group /= lastGroup
            then put (group, rest) >> f ws
            else return $ color grpColor ws

        color grpColor (WSInfo {wsiName=name}) = colorize grpColor "" $ escape name
```

Which colors workspace names in groups according to their prefix
(delimited by '-').

### DIFF
--- a/src/System/Information/EWMHDesktopInfo.hs
+++ b/src/System/Information/EWMHDesktopInfo.hs
@@ -22,6 +22,7 @@
 module System.Information.EWMHDesktopInfo
   ( X11Window      -- re-exported from X11DesktopInfo
   , X11WindowHandle
+  , WorkspaceIdx(..)
   , withDefaultCtx -- re-exported from X11DesktopInfo
   , isWindowUrgent -- re-exported from X11DesktopInfo
   , getCurrentWorkspace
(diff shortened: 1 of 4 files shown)


bgamari:
Finally getting back to this. Your question touches on a concern that I had when I wrote this but never had a chance to return to. In hindsight, I think the solution is pretty straightforward,

Say you expose the `f` as,
```hs
newtype Formatter x a = Formatter (x -> Maybe a)
                        deriving (Functor)

instance Applicative (Formatter x) where
    pure = Formatter . const . Just
    Formatter a <*> Formatter b = Formatter $ \x -> a x <*> b x

instance Alternative (Formatter x) where
    empty = Formatter $ const Nothing
    Formatter a <|> Formatter b = Formatter $ \ws -> a ws <|> b ws

runFormatter :: Formatter x String -> x -> String
runFormatter (Formatter markup) ws
  | Just res <- markup ws = res
  | otherwise             = ""

defaultMarkup :: Formatter WorkspaceInfo Markup
defaultMarkup = Formatter $ Just . go
  where
    go ws@(WSInfo {wsiName=name, wsiVisibility=vis})
      | wsiUrgent ws   = colorize "red" "yellow" $ escape name
      | Active <- vis  = colorize "yellow" "" $ wrap "[" "]" $ escape name
      | Visible <- vis = wrap "(" ")" $ escape name
      | wsiEmpty ws    = escape name
      | otherwise      = escape name
```

But you want to override the style of the `Active` workspace and hide `Empty` workspaces. You could define your style like this,
```hs
markupWorkspaces :: Traversable f => f WorkspaceInfo -> f Markup
markupWorkspaces = fmap (runFormatter $ hideEmpty <|> activeStyle <|> defaultMark)
  where
    hideEmpty = Formatter $ \ws -> Just ""
    activeStyle = ...
```
I still feel like this could be iterated a bit more, however. 

As far as

travitch:
Oh, I really like that.  That is exactly what I had in mind.

I'll try to find some time soon to apply this to your patch and also a few other modules.

IvanMalison:
Closing because WorkspaceSwitcher and Pager have been removed from the codebase. Filed #323 which links here in case we want to revisit this.

PR #102, opened 2014-11-20
Added basic ability to configure the display of MPRIS metadata

pjones:
I added the ability to control the MPRIS display.  This should be directly usable in the MPRIS2 module as well.


jbo-ableton:
Wanting this so much, thanks! :+1:

PR #125, opened 2015-08-10
newtype Workspace indexes

bgamari:
This uses a newtype to represent workspace indexes instead of the current rather unsafe `Int`.


bgamari:
This is split out from #99, which I'll soon try to rebase.

PR #128, opened 2015-10-06
Make battery bar more informative.

saksham0808:
This needs some reworking. As for a single widget having options for both battery bar and text and both. I tried doing that, but that might change some of the previously used things. For example in taffbar.hs, the bar configuration parameter to be passed to the widget would have to be a Maybe now (to allow the user to not pass a barCfg and thus have just a simple text battery), or to pass a configuration and have a bar along with the text.

Also, I'm sure the usage of a Unicode character for the battery charged state would be debatable.

This is just a personal configuration I use. It surely does need some tweaking to be merged.

### DIFF
--- a/src/System/Information/EWMHDesktopInfo.hs
+++ b/src/System/Information/EWMHDesktopInfo.hs
@@ -29,6 +29,7 @@ module System.Information.EWMHDesktopInfo
   , getVisibleWorkspaces
   , getWorkspaceNames
   , switchToWorkspace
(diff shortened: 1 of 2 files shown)


saksham0808:
cabal error while installing dependencies? Sounds weird. Any help?

travitch:
If you are referring to the travis failure, that is a bug in the travis setup.  The custom setup.hs files in the gtk stack cause travis all sorts of problems.

PR #129, opened 2015-10-26
Add support for scrolling on WorkspaceSwitcher

saksham0808:
Scrolling with mouse/touchpad on WorkspaceSwitcher allows moving one workspace up or down.


travitch:
Now that I had a chance to play with this, I notice that the scrolling includes empty desktops.  Is that what you intended?  I'm not sure what I would consider the correct behavior, but I was a bit surprised when scrolling took me to an empty desktop.

saksham0808:
Yes that is what I intended. Though its debatable. I also allowed cyclic scrolling. Your call. Though after a little bit of use, even I feel cycling through non-empty workspaces would be better. I can try cycling through elements of getVisibleWorkspaces.

PR #131, opened 2015-10-29
Fix some issues with taffybar.rc loading

pstray:
This pull request fixes some issues with loading of taffybar.rc, mostly that it is now loaded after initGUI so it can override (and use) any resources loaded by the system gtkrc files, it also fixes an issue with using gtk_color_scheme, which causes already loaded rc files to be reloaded, which could cause overwriting of resources defined in taffybar.rc. It also prefixes all stylesdefined in taffybar.rc with the "taffybar-" so the style names should be more unique.

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -261,15 +261,14 @@ setTaffybarSize cfg window = do
 
 taffybarMain :: TaffybarConfig -> IO ()
 taffybarMain cfg = do
-  -- Override the default GTK theme path settings.  This causes the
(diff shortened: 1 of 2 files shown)


travitch:
Oh, thanks.  I will take a look at this very soon.  I've known that things weren't quite right (especially if you have an XSettings daemon running) for a while, but I never dug into the reasons.  I actually never really understood gtkrc, so this is helpful.

saksham0808:
Not sure if this is the reason. But this would probably make users of light gtk themes hate taffybar. It looks great with a dark theme, but light themes look awful.

PR #147, opened 2016-03-21
Make battery bar more informative.

sakshamsharma:
Had sent this long back. Had been closed since I renamed the branch.
Adds time-remaining and charging/discharging/completed status.

### DIFF
--- a/src/System/Taffybar/Battery.hs
+++ b/src/System/Taffybar/Battery.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 -- | This module provides battery widgets using the UPower system
 -- service.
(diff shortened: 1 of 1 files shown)


sakshamsharma:
@travitch @IvanMalison Can you suggest a better way to do this? At the moment I have to do:
```
cat /sys/class/power_supply/BAT1/status
```
to see the battery. Not something I enjoy. I'm sure adding this feature to master would be a good idea.

IvanMalison:
I can't say that I know much about querying battery status but I have to say that that seems very not portable.

IvanMalison:
@sakshamsharma Can you handle merge conflicts?

sakshamsharma:
Right. had forgotten about this.
Will do soon. I had moved to Nix, where taffybar with GTK3 isn't working somehow, and haven't found the time to debug that.

sakshamsharma:
@IvanMalison I have rebased this PR with the current master, and have tested this code on my system.

My battery widget creation call in taffybar.hs is as follows:
```haskell
batt  = batteryBarNewWithFormat defaultBatteryConfig "$percentage$% ($time$) - $status$" 30
```

There should be no breaking changes, or even usability changes. Also, a particular doc-string's first line had been removed 2 years ago, put it back :smile:

BTW should we also document this new constructor somewhere?

IvanMalison:
> n removed 2 years ago, put it back 
> 
> BTW should we also document this new constructor somewhere?

changelog maybe

sakshamsharma:
Makes sense. Any other issues with this?

Is the unicode character usage okay? It won't be enabled unless the user chooses to.
I would probably like it better if we can pass a configuration flag (boolean) to toggle unicode characters. But that can be done in another PR.

IvanMalison:
> Is the unicode character usage okay? It won't be enabled unless the user chooses to.
> I would probably like it better if we can pass a configuration flag (boolean) to toggle unicode characters. But that can be done in another PR.

The other option would be a function that takes state to markup or something liike that.

IvanMalison:
@sakshamsharma We can iterate it on it in further pull requests. This looks good for now.

sakshamsharma:
Sounds good. Yaay for a new release! Thanks for all your work.

PR #149, opened 2016-05-02
Multiple batteries support

eddsteel:
This changes the battery widget to support multiple batteries. It will display a bar for each battery and a single summarised text description.

- Overall percentage is just the mean, a useful improvement would be to weight it based on individual battery capacity but I get the impression that might not be something we can rely on for all batteries

- Overall time remaining is the sum, if all batteries provide this information. Only one of my two batteries does so I'm not sure if this is the right behaviour generally speaking, but I just use percentage anyway. I would guess this is right for discharging, when one is discharged at a time, and wrong for charging when both are charged simultaneously, maybe someone with two good batteries can confirm. 

Despite these shortcomings I find it useful for my two battery machine.

The first three commits add the new behaviour. I verified the new functions will work for a single battery so the fourth commit removes old functions. Fifth commit renames new functions to match old so that config will still work -- I can drop either or both of these if you want to preserve the old single-battery functions too but I think they're redundant.

### DIFF
--- a/src/System/Information/Battery.hs
+++ b/src/System/Information/Battery.hs
@@ -10,7 +10,6 @@ module System.Information.Battery (
   BatteryTechnology(..),
   BatteryType(..),
   -- * Accessors
-  batteryContextNew,
   batteryContextsNew,
   getBatteryInfo
   ) where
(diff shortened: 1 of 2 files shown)


travitch:
Nice, thanks!

PR #154, opened 2016-06-23
Check names rather than length of workspaces

larkery:
It is possible to rename workspaces using `DynamicWorkspaces`; in this case the list of names changes, without its length changing, and taffybar does not reflect the change. Sometimes this makes taffybar fall over as well, although not always.


travitch:
Thanks - I hadn't thought of dynamic workspaces.

PR #155, opened 2016-06-27
Add configuration option to use the primary display

larkery:
Adds the option to display the bar on the primary xrandr screen rather than a specific screen number.

To avoid breaking backwards compatibility I have added an option to the config rather than making a new type like

```
data Display = Primary | Number Int
```

which would look nicer and allow for future options like maybe `Multiple [1,2]` if anyone wanted to do that. There might be some way to have both by writing a `Num` instance or something but that's a bit much for me.

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -164,6 +164,7 @@ strutProperties pos bh (Rectangle mX mY mW mH) monitors =
 data TaffybarConfig =
   TaffybarConfig { screenNumber :: Int -- ^ The screen number to run the bar on (default is almost always fine)
                  , monitorNumber :: Int -- ^ The xinerama/xrandr monitor number to put the bar on (default: 0)
+                 , primaryMonitor :: Bool -- ^ If true, ignore monitorNumber and use the xrandr primary
                  , barHeight :: Int -- ^ Number of pixels to reserve for the bar (default: 25 pixels)
                  , barPosition :: Position -- ^ The position of the bar on the screen (default: Top)
                  , widgetSpacing :: Int -- ^ The number of pixels between widgets
@@ -177,6 +178,7 @@ defaultTaffybarConfig :: TaffybarConfig
 defaultTaffybarConfig =
   TaffybarConfig { screenNumber = 0
                  , monitorNumber = 0
+                 , primaryMonitor = False
                  , barHeight = 25
                  , barPosition = Top
                  , widgetSpacing = 10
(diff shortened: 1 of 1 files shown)


mellon85:
no progress on this ?

IvanMalison:
@mellon85 There have been a lot of chages to taffybar since this request was made, so this doesn't currently merge.

An effect identical to the one produced by this pull request can be produced by providing the appropriate getMonitorConfig function:

``` haskell
usePrimaryMonitor :: TaffybarConfigEQ -> IO (Int -> Maybe TaffybarConfigEQ)
usePrimaryMonitor c@(cfg, _) = do
  screen <- screenGetPrimaryMonitor $ screenNumber cfg
  primary <- screenGetPrimaryMonitor screen
  return $ (\mnumber -> if mnumber == primary then Just c else Nothing)
```

I wasn't able to test this, but I think it should work. Let me know if it doesn't.

iamjamestl:
@IvanMalison I wasn't able to get your `usePrimaryMonitor` function to work.  I don't see that gtk2hs defines a `screenGetPrimaryMonitor` function.  gi-gdk does, though it's deprecated, and I don't think you can mix gi-gdk and gtk2hs anyway.  Do you have any other ideas?

iamjamestl:
I was able to get this working with the `Graphics.X11.Xrandr` library directly.

```haskell
isActiveOutput :: Display -> XRRScreenResources -> RROutput -> IO Bool
isActiveOutput display sres output = do
    maybeOutputInfo <- xrrGetOutputInfo display sres output
    return $ case maybeOutputInfo of
                Just outputInfo -> (xrr_oi_crtc outputInfo) /= 0
                otherwise       -> False

getActiveOutputs :: Display -> Window -> IO [RROutput]
getActiveOutputs display rootw = do
    maybeSres <- xrrGetScreenResources display rootw
    case maybeSres of
        Just sres ->
            filterM (isActiveOutput display sres) $ xrr_sr_outputs sres
        otherwise ->
            return []

getPrimaryOutputNumber :: ScreenNumber -> IO (Maybe Int)
getPrimaryOutputNumber screenNumber = do
    display <- openDisplay ""
    rootw   <- rootWindow display screenNumber
    primary <- xrrGetOutputPrimary display rootw
    outputs <- getActiveOutputs display rootw
    return $ primary `elemIndex` outputs

usePrimaryMonitor :: TaffybarConfigEQ -> IO (Int -> Maybe TaffybarConfigEQ)
usePrimaryMonitor c@(cfg, _) = do
    maybePrimary <- getPrimaryOutputNumber (fromIntegral $ screenNumber cfg)
    let primary = maybe (monitorNumber cfg) id maybePrimary
    return $ \mnumber -> if mnumber == primary then Just c else Nothing
```

I am new to Haskell, though, so it wouldn't surprise me if this is not very good style.

IvanMalison:
@iamjamestl You might consider packaging that up in to a pull request.

benley:
Should this PR be closed now that #232 is merged?

IvanMalison:
@benley yes

PR #165, opened 2016-11-22
dynamically change colors in vertical bars

teleshoes:
i use it to make my volume bar green-on-black change to red-on-yellow when muted.

### DIFF
--- a/src/System/Taffybar/Widgets/VerticalBar.hs
+++ b/src/System/Taffybar/Widgets/VerticalBar.hs
@@ -8,6 +8,7 @@ module System.Taffybar.Widgets.VerticalBar (
   -- * Accessors/Constructors
   verticalBarNew,
   verticalBarSetPercent,
+  verticalBarSetColors,
   defaultBarConfig
   ) where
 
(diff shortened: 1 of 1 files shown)


travitch:
Just to brainstorm an alternative, what if the type of `barColor` was `Double -> IO (Double, Double, Double)`? Then you would be able to close over an `IORef` or `STMRef`. That is a breaking API change, but the pain of that could be mitigated by adding an alternative `BarConfig` constructor that had the IO versions for `barColor` and `barBackgroundColor`.

In retrospect, I wish I had used the color library there.

teleshoes:
yea, color library....
i like your soln better; its cleaner on your end and more flexible, though posibly messier on mine.

i dont have time to impl it now, tho. any reason not to merge this and then improve it later?

travitch:
I'll take a cut at my solution tonight.  If that doesn't work out, I'll merge your patch

travitch:
Can you take a look at the feature/io-bar-color branch? It adds an alternative constructor with IO variants of the bar configuration options.  I prefer this API, and can do something similar for pager layouts.

teleshoes:
looks great! yea, like you say, the other PR i have open would also be better fixed this way (pager).
thanks!

btw, im working on an implementation of workspace-switcher with icons that uses `_NET_WM_ICON` instead of requiring the user to implement an icon selector for their DE.
this version will have no effect on existing configs unless you add `useImages=True` to pager or something.
if i can make THAT pr acceptable to you, ill FINALLY be able to close my fork. :D

travitch:
Cool!  I'll merge my branch then.  I'll see about doing something similar for your other pull request.

That change you mentioned for the workspace-switcher sounds fine to me.

PR #166, opened 2016-11-24
add activeLayoutIO to pagerConfig

teleshoes:
i use this to add a count of windows in full-screen-mode "[5]"
```
activeLayoutIO   = \layout -> case layout of
      "left"    -> return "[]="
      "top"     -> return "TTT"
      "grid"    -> return "###"
      "full"    -> do
        cnt <- windowCount
        let numFmt = if 0 <= cnt && cnt < 10 then show cnt else "+"
        let color = if cnt > 1 then fgbg "blue" "red" else id
        return $ color $ "[" ++ numFmt ++ "]"
```


teleshoes:
@travitch now that you merged #168, i rebased this one so its no longer broken

PR #167, opened 2016-11-27
workspace images (_NET_WM_ICON and custom files)

teleshoes:
- disabled by default, non-breaking change
- parses EWMH hint _NET_WM_ICON to work out-of-the-box with most modern programs
- allows parsing custom icon files based on the window title and class
  (user can implement a lookup for their desktop environment)

`default config (no images)`
20161126_ss1

`with images enabled`
20161126_ss2

`larger icons, custom image loader for terminal-with-VIM-open`
20161126_ss3

`my config (uses additional GTK code)`
20161126_ss4

PR #168, opened 2016-11-27
workspace styling (workspaceGap, workspacePad, workspaceBorder)

teleshoes:
- all options are disabled by default, non-breaking change
- add an option to pager config to skip pre-pending a space to non-empty workspaces
- add an option to pager config to add a gap in pixels between workspaces
- add an option to pager config to draw a border around workspace buttons
- add gtkrc styles for custom borders in `taffybar.rc`

these options are particularly useful for configuring the workspace images UI

`my config (requires workspace-images code)`
20161126_ss4

### DIFF
--- a/src/System/Taffybar/Pager.hs
+++ b/src/System/Taffybar/Pager.hs
@@ -68,6 +68,7 @@ data PagerConfig = PagerConfig
   , urgentWorkspace  :: String -> String -- ^ workspaces containing windows with the urgency hint set.
   , widgetSep        :: String           -- ^ separator to use between desktop widgets in 'TaffyPager'.
   , workspaceGap     :: Int              -- ^ space in pixels between workspace buttons
+  , workspacePad     :: Bool             -- ^ pad workspace name in button
   , useImages        :: Bool             -- ^ use images in the workspace switcher
   , imageSize        :: Int              -- ^ image height and width in pixels
   , fillEmptyImages  :: Bool             -- ^ fill empty images instead of clearing them
@@ -94,6 +95,7 @@ defaultPagerConfig   = PagerConfig
   , urgentWorkspace  = colorize "red" "yellow" . escape
   , widgetSep        = " : "
   , workspaceGap     = 0
+  , workspacePad     = True
   , useImages        = False
   , imageSize        = 16
   , fillEmptyImages  = False
--- a/src/System/Taffybar/Pager.hs
+++ b/src/System/Taffybar/Pager.hs
@@ -67,6 +67,7 @@ data PagerConfig = PagerConfig
   , visibleWorkspace :: String -> String -- ^ all other visible workspaces (Xinerama or XRandR).
   , urgentWorkspace  :: String -> String -- ^ workspaces containing windows with the urgency hint set.
   , widgetSep        :: String           -- ^ separator to use between desktop widgets in 'TaffyPager'.
+  , workspaceGap     :: Int              -- ^ space in pixels between workspace buttons
   , useImages        :: Bool             -- ^ use images in the workspace switcher
   , imageSize        :: Int              -- ^ image height and width in pixels
   , fillEmptyImages  :: Bool             -- ^ fill empty images instead of clearing them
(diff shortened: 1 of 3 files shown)


teleshoes:
@travitch i rebased and fixed the merge conflicts with #167 (they shared some code) and the warnings fixups. let me know how you feel about this one

IvanMalison:
status?

IvanMalison:
@teleshoes I think it would be nice if icons for all windows in each workspace could be displayed. We would need to change the workspace object to hold an array of images for this to work. any thoughts?

teleshoes:
@IvanMalison what you describe should be possible, but has nothing to do with this PR.
i might work on it in the future, but i think its unlikely to get merged.

this is what you want, more or less, right?

teleshoes:
woo, thanks!

IvanMalison:
> @IvanMalison what you describe should be possible, but has nothing to do with this PR. 

Yeah, I just figured that since you were the one who added the images I'd contact you here.

> i might work on it in the future, but i think its unlikely to get merged.

Hmmm, why?

> this is what you want, more or less, right?

Sort of, I don't really need the grid like grouping of icons. I'd rather just have a contiguous sequence of icons so that every window in the workspace is represented.

My plan was to modify the code so that you could set how many icons were displayed (the rest would be truncated) you could then get the current behavior by setting the number of icons to be displayed to 1.

teleshoes:
by contiguous, you mean arrange them horizontally so that your workspace can be up to twice as wide if you have two windows on every workspace?

heh, sounds awful to me, though im not sure the grid is any better since 12x12 is so small.

im starting to think that this wont be great. perhaps you could mock it up the way you envision it working?

IvanMalison:
> by contiguous, you mean arrange them horizontally so that your workspace can be up to twice as wide if you have two windows on every workspace?

Yep. @teleshoes Why does that sound so awful? I really don't mind how much space the switcher takes up. It seems super useful to be able to know exactly what is in each workspace. In fact, I have something exactly like this set up, except that its implemented with dynamic workspace names in xmonad and fontawesome characters.

teleshoes:
heh, to each their own.

im trying to decide whether it would be better to have `maxRows :: Int` or `useGrid :: Bool`, in addition to `maxIcons`, in order to facilitate both use cases.

so my version would be either `maxIcons=4 ; useGrid=True` or `maxIcons=4 ; maxRows=2`
and yours: `maxIcons=4 ; useGrid=False` or `maxIcons=4 ; maxRows=1`.

maxRows is more general, but the only values that make sense at all are `1` and `2`.
for my bar, maxRows would be 8x8 icons.

no idea when ill impl this btw. im pretty sure i dont want to user it personally, but i might change my mind after i give it a shot.
useGrid is simpler to understand, but slightly more constrictive.

IvanMalison:
@teleshoes does the taffybar.rc theming actually work for you with this? For me it seems to have no effect.

ewolk:
yea. did it work for you before this PR?
if i edit `~/.config/taffybar/taffbar.rc` and restart taffybar, my changes are picked up

teleshoes:
yea. did it work for you before this PR?
if i edit ~/.config/taffybar/taffbar.rc and restart taffybar, my changes are picked up

EDIT: DAMNIT WRONG USER AGAIN

IvanMalison:
Only changes to taffybar-default seem to have any effect for me. I'm unable to make it so that taffybar-workspace-border-active is applied to "*Workspace-*-active*"

teleshoes:
mine is red for active, mustard-yellow otherwise:
```
style "taffybar-workspace-border-active" {
  bg[NORMAL] = "#ff0000"
}
style "taffybar-workspace-border-visible" {
  bg[NORMAL] = "#ff0000"
}
style "taffybar-workspace-border-empty" {
  bg[NORMAL] = "#d4ad35"
}
style "taffybar-workspace-border-hidden" {
  bg[NORMAL] = "#d4ad35"
}
style "taffybar-workspace-border-urgent" {
  bg[NORMAL] = "#d4ad35"
}
```

PR #175, opened 2017-01-31
Add support for multiple icon image display

IvanMalison:
@teleshoes As discussed in https://github.com/travitch/taffybar/pull/168 I've added support for multiple image display. I've set the default behavior to be to display up to 4 icons, but I'm open to changing this default to 1, but I think that having multiple icons is much more useful.


src/System/Taffybar/Pager.hs
```diff
@@ -102,6 +104,7 @@ defaultPagerConfig   = PagerConfig
   , fillEmptyImages  = False
   , preferCustomIcon = False
   , customIcon       = \_ _ -> Nothing
+  , imageCount       = 4
```
teleshoes: for my taffybar, 4 icons completely fills my taffybar, with nothing but wmlog.
(my wmlog is ~20% of my bar already)

IvanMalison:
Here's what this looks like now: http://imgur.com/a/Ec6Jg

IvanMalison:
Superceded by #180

Issue #174, opened 2017-01-30
serious performance regression for _NET_WM_DESKTOP


teleshoes:
noticed that switching workspaces was massively sluggish after i rebased on the latest.

i bisected the regression to `1a5c422`.

the workspace-switcher now takes about a full second to update. its so slow now that by the time it updates, im somewhere else.

if i switch back and forth between two workspaces about 20 times, the workspace-switcher freezes and my cpu spikes for ~70s.

perhaps this should be added as a configuration option to fix i3wm

@sakshamsharma @travitch

teleshoes:
1) actually, the whole taffybar freezes
2) the regression only happens with `useImages=True`

IvanMalison:
@teleshoes A _NET_WM_DESKTOP callback is still needed to appropriately set the icon if the icon that you are interested in is not on the current workspace, so it seems to me that it DOES make sense to listen on this callback. It seems likely that what is slow is the actual setting of images itself. It might make sense to start doing some caching for the icons (i.e. only set the pixelbuffer for the image if the associate icon IS actually changed)

ewolk:
1) actually, _NET_WM_DESKTOP is not necessary for setting icons on non-active workspaces
2) yea, the setting of the icons is what takes cpu time, but this regression is because the _NET_WM_DESKTOP callback code is resulting in it being called many many times when the workspace changes
3) the solution is to separate out exactly what needs updating on what events. this isnt quite trivial to do. IMO, a good interim solution would be to add a config option to disable listening to _NET_WM_DESKTOP
4) yea, i experimented with caching pixbufs
  a) you cannot cache by window title and window class, as the icon for an app can change even while the program is still running, let alone run-to-run
  b) the performance was more or less identical
  c) the caching code was fairly complicated on my first pass, and might be a burder to maintain in future, i think.
  d) i truly dont think caching is an improvement, as **its quite fast as long as you dont call it 64 times when you switch workspaces**

ewolk:
tested with unsafePerformIO in activecb with 10 windows in total, spread over 5 workspaces:
1) _NET_WM_DESKTOP is called 4-15 times when i switch windows within a workspace
2) _NET_WM_DESKTOP is called 8-25 times when i switch workspaces
3) _NET_WM_DESKTOP is called exactly 7 times when i push enter in urxvt

(all those things result in a single _NET_CURRENT_DESKTOP event)

[...]

PR #176, opened 2017-01-31
_NET_WM_DESKTOP performance regression

teleshoes:
fixes regression with `useImage=True` caused by `_NET_WM_DESKTOP` event for i3wm #174


teleshoes:
this PR is minimalistic. i would personally have preferred to refactor the event code to:
1) have more accurate comments (many things have changed)
and
2) name the callbacks based on the events they handle
they are currently SORT OF named instead after what they do, but for the `active` callback at least, these days it does so much more than just setting the active desktop.
(IMO, the callbacks should call other functions that do explicit things like mark-urgent where appropriate)

but i want to make as little change as possible to fix the regression
let me know what you think
@travitch

teleshoes:
this is no longer necessary

Issue #178, opened 2017-02-07
stack ghci does not work


IvanMalison:
stack ghci fails with the following output:
```
/usr/bin/ld: /home/imalison/Projects/taffybar/.stack-work/dist/x86_64-linux-tinfo6/Cabal-1.22.5.0/build/
taffybar/taffybar-tmp/src/gdk_property_change_wrapper.o: relocation R_X86_64_32 against `.rodata.str1.1'
 can not be used when making a shared object; recompile with -fPIC                                     
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
```

IvanMalison:
This is now fixed as of ccbf7e45f76781827420f5cf8ff26a26da2eba80

IvanMalison:
If you get this error delete the .stack-work directory and try again.

PR #164, opened 2016-10-20
Build with GHC 8

sakshamsharma:
Please test build with `cabal` before merging. I don't use it anymore.
Although I suppose it should build fine.

Also, since the build had stopped working thanks to [this issue](https://github.com/gtk2hs/gtk2hs/issues/179#issuecomment-255053104), I have added the required package to the cabal file. Seems to build on my system ATM, need more people to confirm.

Lastly, I've added some comments on installation in the README.md as well. Please refer to the issue linked above for details.

Issue #105, opened 2014-12-24
Crash when multiple windows close simultaneously


ertes:
Hi there,

Taffybar crashes with [my configuration](http://hub.darcs.net/ertes/config/browse/mytaffybar) when two windows are destroyed simultaneously. This is how I can reproduce it:
- Run: urxvt -e screen -S test
- Run: urxvt -e screen -x test
- In one of those windows exit the shell.

This causes the screen session to end and both windows to close at the same time. Taffybar goes with them, resulting in the following error:

```
% mytaffybar
Monitor 0 is not available in the selected screen
The program 'mytaffybar' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 12 error_code 3 request_code 20 minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

I don't mind restarting Taffybar as nothing really depends on it, but I close multiple windows at the same time very often (browser windows, a set of pictures, Gimp windows, etc.).

travitch:
Can you try the latest on branch issue-105?  It is more systemic in ensuring that GUI updates are make safely via postGUIAsync.  I am reasonably sure that there were at least a few calls being made in another thread before; that could have caused an issue.

Nadrieril:
I'm having this too (my config: https://github.com/Nadrieril/taffybar-config)

jagajaga:
Having this issue now :(
```
The program 'taffybar-linux-x86_64' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 21 error_code 3 request_code 20 minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

simlu:
Running into this as well. Seems to happen only when:
* I use `wspaceSwitcherNew`, `layoutSwitcherNew` or `windowSwitcherNew` pager
* Close and open windows

Let me know how to provide more debugging information...

Result of running `taffybar --sync` below:
```
[...]

PR #180, opened 2017-02-17
WorkspaceHUD

### LINKED ISSUE #174: serious performance regression for _NET_WM_DESKTOP

IvanMalison:
This pull request is essentially a complete rewrite of WorkspaceSwitcher that seeks to accomplish a few different things:

- Separate the business logic and UI code of the workspace switcher
- Provide an interface to allow advanced users to create completely bespoke workspace widgets (or compose their own UI elements with existing UI code). See WorkspaceWidgetController for details about how this works.
- Drastically improve image performance -- Because each workspace contents widget keeps the last Workspace model that it saw around, it can be intelligent about when it should update the workspace icons (only when something has changed in the workspace, which is not often). Also because this is done on a per workspace basis, even when something does change, we only redraw the images for that particular workspace. This approach is a much cleaner solution to #174 than #176.
- Display of multiple images -- the aforementioned performance improvements make it feasible to display an icon for every window in each workspace. This is much more useful than having a single icon per workspace because the user can see where every window they have open is.
- Better urgency handling -- Beacuse the HUD displays an image for every window, we can highlight the icon of the window that is set to urgent, which is obviously better than simply highlighting the entire workspace.
- Clicking on icons now focuses on the associated window, not just the workspace that contains it. Not that useful, but kinda cool.
- Better state indication -- The colored [underlines](http://imgur.com/a/bC2x9) look pretty sweet, don't they? These are themed using taffybar.rc instead of through pagerConfig
- Cleaner code -- obviously this is just my opinion, but I think that WorkspaceHUD's code is much closer to what one might describe as idiomatic haskell than what is currently in WorkspaceSwitcher.hs. WorkspaceSwitcher uses tuple types where ADTs are appropriate and case statements where <$>, >>= etc. work much more nicely.

Anyway, I did all of this without doing anything to disrupt WorkspaceSwitcher so this should be a relatively harmless merge. As I mentioned before, I would like this to completely supercede WorkspaceSwitcher, but I understand that other people might feel differently.

@travitch, @teleshoes, @escherdragon @sakshamsharma -- let me know what you think, and if any of you are likely to use any of this. My taffybar config lives [here](https://github.com/IvanMalison/dotfiles/tree/master/dotfiles/config/taffybar) if you want to get something that looks like what is show in the images I provided.

http://imgur.com/a/bC2x9

Fixes #174, #178

### DIFF
--- a/src/System/Taffybar/IconImages.hs
+++ b/src/System/Taffybar/IconImages.hs
@@ -10,9 +10,11 @@
 -----------------------------------------------------------------------------
 
 module System.Taffybar.IconImages (
+  ColorRGBA,
   scalePixbuf,
   pixBufFromEWMHIcon,
   pixelsARGBToBytesRGBA,
+  pixBufFromColor,
   pixBufFromFile,
   selectEWMHIcon
 ) where
@@ -24,6 +26,8 @@ import           Foreign.C.Types (CUChar(..))
 import           Foreign.Marshal.Array (newArray)
 import qualified Graphics.UI.Gtk as Gtk
 import           System.Information.EWMHDesktopInfo
+
+type ColorRGBA = (Word8, Word8, Word8, Word8)
 
 -- | Take the passed in pixbuf and ensure its scaled square.
 scalePixbuf :: Int -> Gtk.Pixbuf -> IO Gtk.Pixbuf
@@ -49,6 +53,17 @@ pixBufFromEWMHIcon EWMHIcon {width=w, height=h, pixelsARGB=px} = do
       bytesRGBA = pixelsARGBToBytesRGBA px
   cPtr <- newArray $ map CUChar bytesRGBA
   Gtk.pixbufNewFromData cPtr colorspace hasAlpha sampleBits w h rowStride
+
+-- | Create a pixbuf with the indicated RGBA color,
+-- scale it square, and set it in a GTK Image.
+pixBufFromColor :: Int -> ColorRGBA -> IO Gtk.Pixbuf
+pixBufFromColor imgSize (r,g,b,a) = do
+  let sampleBits = 8
+      hasAlpha = True
+      colorspace = Gtk.ColorspaceRgb
+  pixbuf <- Gtk.pixbufNew colorspace hasAlpha sampleBits imgSize imgSize
+  Gtk.pixbufFill pixbuf r g b a
+  return pixbuf
 
 -- | Convert a list of integer pixels to a bytestream with 4 channels.
 pixelsARGBToBytesRGBA :: [Int] -> [Word8]
--- a/src/System/Taffybar/IconImages.hs
+++ b/src/System/Taffybar/IconImages.hs
@@ -1,0 +1,73 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      : System.Taffybar.IconImages
+-- Copyright   : (c) Ivan A. Malison
+-- License     : BSD3-style (see LICENSE)
+--
+-- Maintainer  : Ivan A. Malison
+-- Stability   : unstable
+-- Portability : unportable
+-----------------------------------------------------------------------------
+
+module System.Taffybar.IconImages (
+  scalePixbuf,
+  pixBufFromEWMHIcon,
+  pixelsARGBToBytesRGBA,
+  pixBufFromFile,
+  selectEWMHIcon
+) where
+
+import           Data.List
+import           Data.Ord
+import           Data.Word (Word8)
+import           Foreign.C.Types (CUChar(..))
+import           Foreign.Marshal.Array (newArray)
+import qualified Graphics.UI.Gtk as Gtk
+import           System.Information.EWMHDesktopInfo
+
+-- | Take the passed in pixbuf and ensure its scaled square.
+scalePixbuf :: Int -> Gtk.Pixbuf -> IO Gtk.Pixbuf
+scalePixbuf imgSize pixbuf = do
+  h <- Gtk.pixbufGetHeight pixbuf
+  w <- Gtk.pixbufGetWidth pixbuf
+  if h /= imgSize || w /= imgSize
+  then
+    Gtk.pixbufScaleSimple pixbuf imgSize imgSize Gtk.InterpBilinear
+  else
+    return pixbuf
+
+-- | Create a pixbuf from the pixel data in an EWMHIcon,
+-- scale it square, and set it in a GTK Image.
+pixBufFromEWMHIcon :: EWMHIcon -> IO Gtk.Pixbuf
+pixBufFromEWMHIcon EWMHIcon {width=w, height=h, pixelsARGB=px} = do
+  let pixelsPerRow = w
+      bytesPerPixel = 4
+      rowStride = pixelsPerRow * bytesPerPixel
+      sampleBits = 8
+      hasAlpha = True
+      colorspace = Gtk.ColorspaceRgb
+      bytesRGBA = pixelsARGBToBytesRGBA px
+  cPtr <- newArray $ map CUChar bytesRGBA
+  Gtk.pixbufNewFromData cPtr colorspace hasAlpha sampleBits w h rowStride
+
+-- | Convert a list of integer pixels to a bytestream with 4 channels.
+pixelsARGBToBytesRGBA :: [Int] -> [Word8]
+pixelsARGBToBytesRGBA (x:xs) = r:g:b:a:pixelsARGBToBytesRGBA xs
+  where r = toByte $ x `div` 0x10000   `mod` 0x100
+        g = toByte $ x `div` 0x100     `mod` 0x100
+        b = toByte $ x                 `mod` 0x100
+        a = toByte $ x `div` 0x1000000 `mod` 0x100
+        toByte i = (fromIntegral i) :: Word8
+pixelsARGBToBytesRGBA _ = []
+
+-- | Create a pixbuf from a file,
+-- scale it square, and set it in a GTK Image.
+pixBufFromFile :: Int -> FilePath -> IO Gtk.Pixbuf
+pixBufFromFile imgSize file = Gtk.pixbufNewFromFileAtScale file imgSize imgSize False
+
+selectEWMHIcon :: Int -> [EWMHIcon] -> EWMHIcon
+selectEWMHIcon imgSize icons = head prefIcon
+  where sortedIcons = sortBy (comparing height) icons
+        smallestLargerIcon = take 1 $ dropWhile ((<= imgSize) . height) sortedIcons
+        largestIcon = take 1 $ reverse sortedIcons
+        prefIcon = smallestLargerIcon ++ largestIcon
--- a/src/System/Taffybar/IconImages.hs
+++ b/src/System/Taffybar/IconImages.hs
@@ -43,7 +43,7 @@ scalePixbuf imgSize pixbuf = do
 -- | Create a pixbuf from the pixel data in an EWMHIcon,
 -- scale it square, and set it in a GTK Image.
 pixBufFromEWMHIcon :: EWMHIcon -> IO Gtk.Pixbuf
-pixBufFromEWMHIcon EWMHIcon {width=w, height=h, pixelsARGB=px} = do
+pixBufFromEWMHIcon EWMHIcon {width = w, height = h, pixelsARGB = px} = do
   let pixelsPerRow = w
       bytesPerPixel = 4
       rowStride = pixelsPerRow * bytesPerPixel
@@ -57,7 +57,7 @@ pixBufFromEWMHIcon EWMHIcon {width=w, height=h, pixelsARGB=px} = do
 -- | Create a pixbuf with the indicated RGBA color,
 -- scale it square, and set it in a GTK Image.
 pixBufFromColor :: Int -> ColorRGBA -> IO Gtk.Pixbuf
-pixBufFromColor imgSize (r,g,b,a) = do
+pixBufFromColor imgSize (r, g, b, a) = do
   let sampleBits = 8
       hasAlpha = True
       colorspace = Gtk.ColorspaceRgb
--- a/src/System/Taffybar/WorkspaceHUD.hs
+++ b/src/System/Taffybar/WorkspaceHUD.hs
@@ -228,7 +228,7 @@ updateWorkspaceControllers c@Context { controllersVar = controllersRef
         removeWorkspaces = (Set.difference existingWorkspacesSet newWorkspacesSet)
         builder = (widgetBuilder cfg) cfg
     MV.modifyMVar_ controllersRef $ \controllers -> do
-      let oldRemoved = foldl (flip M.delete) controllers removeWorkspaces
+      let oldRemoved = F.foldl (flip M.delete) controllers removeWorkspaces
           buildController idx =
               case (M.lookup idx workspacesMap) of
                 Just ws -> builder ws
--- a/src/System/Taffybar/WorkspaceHUD.hs
+++ b/src/System/Taffybar/WorkspaceHUD.hs
@@ -19,6 +19,7 @@ module System.Taffybar.WorkspaceHUD (
   WorkspaceUnderlineController(..),
   WorkspaceHUDConfig(..),
   WorkspaceWidgetController(..),
+  buildBorderButtonController,
   buildButtonController,
   buildContentsController,
   buildUnderlineButtonController,
@@ -110,6 +111,7 @@ data WorkspaceHUDConfig =
   , labelSetter :: Workspace -> String
   , updateIconsOnTitleChange :: Bool
   , showWorkspaceFn :: Workspace -> Bool
+  , borderWidth :: Int
   }
 
 hudFromPagerConfig :: PagerConfig -> WorkspaceHUDConfig
@@ -142,6 +144,7 @@ defaultWorkspaceHUDConfig =
                      , labelSetter = workspaceName
                      , updateIconsOnTitleChange = True
                      , showWorkspaceFn = const True
+                     , borderWidth = 2
                      }
 
 hideEmpty :: Workspace -> Bool
@@ -616,11 +619,11 @@ instance WorkspaceWidgetController WorkspaceButtonController
 
 data WorkspaceUnderlineController =
   WorkspaceUnderlineController { table :: T.Table
-                      -- XXX: An event box is used here because we need to
-                      -- change the background
-                      , underline :: Gtk.EventBox
-                      , overlineController :: WWC
-                      }
+                               -- XXX: An event box is used here because we need to
+                               -- change the background
+                               , underline :: Gtk.EventBox
+                               , overlineController :: WWC
+                               }
 
 buildUnderlineController :: ParentControllerConstructor
 buildUnderlineController contentsBuilder context workspace = do
@@ -652,6 +655,40 @@ updateUnderline uc u = do
   newContents <- updateWidget (overlineController uc) u
   return uc { overlineController = newContents }
 
+data WorkspaceBorderController =
+  WorkspaceBorderController { border :: Gtk.EventBox
+                            , borderContents :: Gtk.EventBox
+                            , insideController :: WWC
+                            }
+
+buildBorderController :: ParentControllerConstructor
+buildBorderController contentsBuilder context workspace = do
+  cc <- contentsBuilder context workspace
+  brd <- Gtk.eventBoxNew
+  cnt <- Gtk.eventBoxNew
+  Gtk.eventBoxSetVisibleWindow brd True
+  Gtk.containerSetBorderWidth cnt (borderWidth $ hudConfig context)
+  Gtk.containerAdd brd cnt
+  Gtk.containerAdd cnt $ getWidget cc
+  return $ WWC $ WorkspaceBorderController { border = brd
+                                           , borderContents = cnt
+                                           , insideController = cc
+                                           }
+
+instance WorkspaceWidgetController WorkspaceBorderController
+  where
+    getWidget bc = Gtk.toWidget $ border bc
+    updateWidget bc wu@(WorkspaceUpdate workspace) =
+      (Gtk.widgetSetName (border bc) $ getWidgetName workspace "Border") >>
+      (Gtk.widgetSetName (borderContents bc) $ getWidgetName workspace "Container") >>
+      (updateBorder bc wu)
+    updateWidget a b = updateBorder a b
+
+updateBorder :: WorkspaceBorderController -> WidgetUpdate -> IO WorkspaceBorderController
+updateBorder bc update = do
+  newContents <- updateWidget (insideController bc) update
+  return bc { insideController = newContents }
+
 getWidgetName :: Workspace -> String -> String
 getWidgetName ws wname =
   printf
@@ -663,3 +700,7 @@ getWidgetName ws wname =
 buildUnderlineButtonController :: ControllerConstructor
 buildUnderlineButtonController =
   buildButtonController (buildUnderlineController buildContentsController)
+
+buildBorderButtonController :: ControllerConstructor
+buildBorderButtonController =
+  buildButtonController (buildBorderController buildContentsController)
--- a/src/System/Taffybar/WorkspaceHUD.hs
+++ b/src/System/Taffybar/WorkspaceHUD.hs
@@ -119,7 +119,7 @@ data WorkspaceHUDConfig =
   , borderWidth :: Int
   , updateEvents :: [String]
   , updateRateLimitMicroseconds :: Integer
-  , logUpdates :: Bool
+  , debugMode :: Bool
   }
 
 hudFromPagerConfig :: PagerConfig -> WorkspaceHUDConfig
@@ -186,7 +186,7 @@ defaultWorkspaceHUDConfig =
       , "WM_HINTS"
       ]
   , updateRateLimitMicroseconds = 100000
-  , logUpdates = False
+  , debugMode = False
   }
 
 hideEmpty :: Workspace -> Bool
@@ -196,6 +196,7 @@ hideEmpty _ = True
 data Context =
   Context { controllersVar :: MV.MVar (M.Map WorkspaceIdx WWC)
           , workspacesVar :: MV.MVar (M.Map WorkspaceIdx Workspace)
+          , loggingVar :: MV.MVar Bool
           , hudWidget :: Gtk.HBox
           , hudConfig :: WorkspaceHUDConfig
           , hudPager :: Pager
@@ -203,7 +204,9 @@ data Context =
 
 
 hudLogger :: Context -> String -> IO ()
-hudLogger context = when (logUpdates $ hudConfig context) . putStrLn
+hudLogger Context { loggingVar = loggingRef} txt = do
+  shouldLog <- MV.readMVar loggingRef
+  when shouldLog $ putStrLn txt
 
 updateVar :: MV.MVar a -> (a -> IO a) -> IO a
 updateVar var modify = MV.modifyMVar var $ fmap (\a -> (a, a)) . modify
@@ -261,14 +264,18 @@ buildWorkspaces _ = do
                               , windows = windowInfos
                               } theMap) M.empty names
 
+
 addWidgetsToTopLevel :: Context -> IO ()
-addWidgetsToTopLevel Context { controllersVar = controllersRef
-                             , hudWidget = cont
-                             } = do
+addWidgetsToTopLevel c@Context { controllersVar = controllersRef
+                               , loggingVar = loggingRef
+                               , hudWidget = cont
+                               , hudConfig = cfg
+                               } = do
   controllersMap <- MV.readMVar controllersRef
   -- Elems returns elements in ascending order of their keys so this will always
   -- add the widgets in the correct order
   mapM_ addWidget $ M.elems controllersMap
+  when (debugMode cfg) addDebugWidgets
   Gtk.widgetShowAll cont
     where addWidget controller = do
             -- XXX: This hbox exists to (hopefully) prevent the issue where
@@ -278,14 +285,42 @@ addWidgetsToTopLevel Context { controllersVar = controllersRef
             Gtk.widgetReparent (getWidget controller) hbox
             Gtk.containerAdd hbox $ getWidget controller
             Gtk.containerAdd cont hbox
+          addDebugWidgets = do
+            enableLoggingBox <- Gtk.eventBoxNew
+            rebuildBarBox <- Gtk.eventBoxNew
+            logLabel <- Gtk.labelNew $ Just "ToggleLogging"
+            rebuildLabel <- Gtk.labelNew $ Just "Rebuild Bar"
+            let toggleLogging = MV.modifyMVar_ loggingRef
+                                (\current -> do
+                                   let newState = not current
+                                       labelText = printf "ToggleLogging: %s" $
+                                         show newState :: String
+                                   Gtk.labelSetMarkup logLabel labelText
+                                   return $ not current) >> return True
+                rebuildBar =
+                  do
+                    -- Clear the container and repopulate it
+                    Gtk.containerForeach cont (Gtk.containerRemove cont)
+                    addWidgetsToTopLevel c
+                    updateAllWorkspaceWidgets c
+                    return True
+            Gtk.containerAdd enableLoggingBox logLabel
+            Gtk.containerAdd rebuildBarBox rebuildLabel
+            _ <- Gtk.on enableLoggingBox Gtk.buttonPressEvent (liftIO toggleLogging)
+            _ <- Gtk.on rebuildBarBox Gtk.buttonPressEvent (liftIO rebuildBar)
+            Gtk.containerAdd cont enableLoggingBox
+            Gtk.containerAdd cont rebuildBarBox
+            return ()
 
 buildWorkspaceHUD :: WorkspaceHUDConfig -> Pager -> IO Gtk.Widget
 buildWorkspaceHUD cfg pager = do
   cont <- Gtk.hBoxNew False (widgetGap cfg)
   controllersRef <- MV.newMVar M.empty
   workspacesRef <- MV.newMVar M.empty
+  loggingRef <- MV.newMVar False
   let context = Context { controllersVar = controllersRef
                         , workspacesVar = workspacesRef
+                        , loggingVar = loggingRef
                         , hudWidget = cont
                         , hudConfig = cfg
                         , hudPager = pager
--- a/src/System/Taffybar/WorkspaceHUD.hs
+++ b/src/System/Taffybar/WorkspaceHUD.hs
@@ -215,10 +215,15 @@ onActiveChanged context _ =
 
 
 
+data IconWidget = IconWidget { iconContainer :: Gtk.EventBox
+                             , iconImage :: Gtk.Image
+                             , iconWindow :: MV.MVar X11Window
+                             }
+
 data WorkspaceContentsController = WorkspaceContentsController
   { container :: Gtk.HBox
   , label :: Gtk.Label
-  , iconImages :: [Gtk.Image]
+  , iconImages :: [IconWidget]
   , contentsWorkspace :: Workspace
   , contentsConfig :: WorkspaceHUDConfig
   }
@@ -250,7 +255,8 @@ instance WorkspaceWidgetController WorkspaceContentsController where
          Gtk.labelSetMarkup (label cc) (workspaceName newWorkspace)
 
     newImages <-
-      if ((windowIds currentWorkspace) /= (windowIds newWorkspace))
+      if ((windowIds currentWorkspace) /= (windowIds newWorkspace) ||
+          (urgentIds currentWorkspace) /= (urgentIds newWorkspace))
       then
         updateImages cc newWorkspace
       else
@@ -279,33 +285,37 @@ getIconInfo cfg w = do
            then IINone
            else IIEWMH $ selectEWMHIcon (windowIconSize cfg) icons
 
-updateImages :: WorkspaceContentsController -> Workspace -> IO [Gtk.Image]
+forkM :: Monad m => (i -> m a) -> (i -> m b) -> i -> m (a, b)
+forkM a b input = do
+  resA <- a input
+  resB <- b input
+  return (resA, resB)
+
+updateImages :: WorkspaceContentsController -> Workspace -> IO [IconWidget]
 updateImages wcc ws = do
-  iconInfos_ <- mapM (getIconInfo (contentsConfig wcc)) $ windowIds ws
+  iconInfos_ <- mapM (forkM (getIconInfo $ contentsConfig wcc) return) $ windowIds ws
   -- XXX: Only one of the two things being zipped can be an infinite list, which
   -- is why this newImagesNeeded contortion is needed.
   let iconInfos =
         if newImagesNeeded
           then iconInfos_
-          else (iconInfos_ ++ repeat IINone)
+          else (iconInfos_ ++ repeat (IINone, 0))
 
-  newImgs <- zipWithM setImageFromIO getImgs iconInfos
+  newImgs <- zipWithM updateIconWidget' getImgs iconInfos
   when newImagesNeeded $ Gtk.widgetShowAll $ container wcc
   return newImgs
 
   where
-    imgSize = windowIconSize $ contentsConfig wcc
-    preferCustom = False
-    setImageFromIO getImage iconInfo = do
-      img <- getImage
-      setImage imgSize preferCustom img iconInfo
-      return img
+    updateIconWidget' getImage (iconInfo, windowId)  = do
+      iconWidget <- getImage
+      _ <- updateIconWidget wcc ws iconWidget iconInfo windowId
+      return iconWidget
     infiniteImages =
       (map return $ iconImages wcc) ++
       (repeat $ do
-         img <- Gtk.imageNew
-         Gtk.containerAdd (container wcc) img
-         return img)
+         iw <- buildIconWidget
+         Gtk.containerAdd (container wcc) $ iconContainer iw
+         return iw)
     newImagesNeeded = (length $ iconImages wcc) < (length $ windowIds ws)
     imgSrcs =
       if newImagesNeeded
@@ -314,6 +324,47 @@ updateImages wcc ws = do
     getImgs = case maxIcons $ contentsConfig wcc of
                 Just theMax -> take theMax imgSrcs
                 Nothing -> imgSrcs
+
+buildIconWidget :: IO IconWidget
+buildIconWidget = do
+  img <- Gtk.imageNew
+  ebox <- Gtk.eventBoxNew
+  windowVar <- MV.newMVar 0
+  Gtk.containerAdd ebox img
+  _ <- Gtk.on ebox Gtk.buttonPressEvent $ liftIO $ do
+                    window <- MV.readMVar windowVar
+                    withDefaultCtx $ focusWindow window
+                    return True
+  return IconWidget { iconContainer = ebox
+                    , iconImage = img
+                    , iconWindow = windowVar
+                    }
+
+updateIconWidget
+  :: WorkspaceContentsController
+  -> Workspace
+  -> IconWidget
+  -> IconInfo
+  -> X11Window
+  -> IO IconWidget
+updateIconWidget wcc ws iw@IconWidget { iconContainer = iconButton
+                                      , iconImage = image
+                                      , iconWindow = windowRef
+                                      } info windowId  =
+  do
+    let imgSize = windowIconSize $ contentsConfig wcc
+        preferCustom = False
+        urgent = elem windowId $ urgentIds ws
+        urgentStr = if urgent then "urgent" else "normal"
+
+    setImage imgSize preferCustom image info
+
+    let widgetName = printf "Workspace-icon-%s-%s" (show windowId) urgentStr
+    Gtk.widgetSetName iconButton (widgetName :: String)
+
+    MV.modifyMVar_ windowRef $ const $ return windowId
+
+    return iw
 
 -- | Sets an image based on the image choice (EWMHIcon, custom file, and fill color).
 setImage :: Int -> Bool -> Gtk.Image -> IconInfo -> IO ()
@@ -415,7 +466,6 @@ buildUnderlineButtonController =
   buildButtonController (buildUnderlineController buildContentsController)
 
 -- TODO:
--- * Handle urgent
 -- * Custom Files/Colors
 -- * Handle redraw for new workspace
 -- * scoll on button for switch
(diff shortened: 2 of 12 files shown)


teleshoes:
lot to take in here
1) the UI definitely looks a lot sexier, but IMO its less useful. this is the main stuff i dont like:
-i dislike that it has **variable width** (plus its too wide in general). i personally like to have fixed locations on screen. (if it can be configured to have exactly one icon with an empty-placeholder, that would be enough for me)
-i think the urgency state isnt urgent-looking-enough. i prefer changing the bgcolor of the whole button
-i have some trouble with the colors you picked for the underlining, and because i use **smaller buttons** for each workspace, i think the **underlines are too understated** to actually be useful (i prefer the border PLUS the font color change, a lot more clear i think)

2) all this would be ok as the default if there were easy alternatives, but it seems you **removed most of the customizability** that WSS has. i really like how you **expose the gtk widgets**, but IMO its much less useful than the simple methods. (all-in-all, this feels like a framework instead of a toolkit. maybe it looks nicer, but IMO the point of taffybar is to build your own systembar).
users who want a simple, text-only WSS are in trouble. users with colorblindness would need to code their own wrapper to indicate urgency with text or different colors, etc etc

3) this is a small feature removal: windows can and do change their icons at runtime. this looks like you dont update the icons unless the list of windows is different. like i said, this is not terribly important, but some apps have a startup-icon that will stay in place until the user rearranges the windows, so it will be a bit buggy.

overall, good job on making something that looks so cool. some users would definitely prefer it, but i think **this will never replace WSS for me personally**.
i think the biggest difference between your use case and mine is that this is essentially the only thing you use taffybar for, and i have 26 other widgets, so i need a more concise (if less sexy) WSS

IMO, **some features would be excellent additions to WSS, though.** exposing the gtk widgets, the cleaner icon handling, the ability to have multiple icons (not for me but still cool), etc.
i selfishly want you to reconsider doing your first approach and modifying WSS without removing features, and then adding the rest of this code to your own taffybar using your workspace-widget-controller-api

p.s.: MY WORKSPACE BORDERS DONT WORK ON MY OLD LAPTOP.
just like you said, the styles dont affect the frame. this is really odd, because all the documentation says you CANT set GtkFrame background colors, but here i am, looking at it working on one computer, and broken on another with identical versions of gtk2hs, ghc, gtk, X11, and taffybar.
must be some library difference somewhere, but i really cant see what. just thought i should mention.

IvanMalison:
> -i dislike that it has variable width (plus its too wide in general). i personally like to have fixed locations on screen. (if it can be configured to have exactly one icon with an empty-placeholder, that would be enough for me)

Exactly one icon is possible. Non variable width would also be easy. There is a configuration option to set minimum width. It would be easy to add one to set maximum width as well.

> -i think the urgency state isnt urgent-looking-enough. i prefer changing the bgcolor of the whole button
> -i have some trouble with the colors you picked for the underlining, and because i use smaller buttons for each workspace, i think the underlines are too understated to actually be useful (i prefer the border PLUS the font color change, a lot more clear i think)

This can be accomplished with a different taffybar.rc . All that needs to be done is to have the contents controller set the widget name of the label. This should not be done in haskell for the sake of consistency.

> all this would be ok as the default if there were easy alternatives, but it seems you removed most of the customizability that WSS has. i really like how you expose the gtk widgets, but IMO its much less useful than the simple methods. (all-in-all, this feels like a framework instead of a toolkit. maybe it looks nicer, but IMO the point of taffybar is to build your own systembar).

I think, you are kind of missing the point here. WorkspaceHUD has much MORE customizability than WSS because you can specify your own UI builder. That is precisely why I built this, because I didn't like the way the defaults work.

> users who want a simple, text-only WSS are in trouble. users with colorblindness would need to code their own wrapper to indicate urgency with text or different colors, etc etc

I'm not convinced that this is the case. You can get text only by setting maxImages to 0.

The colors are set in taffybar.rc, so you can choose whatever colors you want.

I'll add a function that allows the user to specify how they want to transform the label text given the workspace object. That will allow at least as much flexibility as WSS has with respect to the label.

> IMO, some features would be excellent additions to WSS, though. exposing the gtk widgets, the cleaner icon handling, the ability to have multiple icons (not for me but still cool), etc.
> i selfishly want you to reconsider doing your first approach and modifying WSS without removing features, and then adding the rest of this code to your own taffybar using your workspace-widget-controller-api

The WSS code is really untidy IMO.

How about this:
Send me your exact configuration of WSS, and I will replicate it in WorkspaceHUD, with better performance.

src/System/Taffybar/WorkspaceHUD.hs
```diff
@@ -0,0 +1,504 @@
...
+            wClass <- withDefaultCtx $ getWindowClass w
+            let customResult = customIconF wTitle wClass
+            defaultResult <- defaultGetIconInfo cfg w
+            let first = if preferCustom then customResult else defaultResult
+            let second = if preferCustom then defaultResult else customResult
+            return $ case first of
+                       IINone -> second
+                       _ -> first
+
+forkM :: Monad m => (i -> m a) -> (i -> m b) -> i -> m (a, b)
```
teleshoes: `forkM` sounds erroneously like concurrency.
in pointless-haskell its called `/|\` and is described as the `monadic split combinator`.
id say `splitM` or even something like `applyZipM` would be better.

src/System/Taffybar/WorkspaceHUD.hs
```diff
@@ -0,0 +1,504 @@
...
+forkM = ((liftM2 . liftM2) (,))
+
+updateImages :: WorkspaceContentsController -> Workspace -> IO [IconWidget]
+updateImages wcc ws = do
+  let cfg = contentsConfig wcc
+  iconInfos_ <- mapM (forkM (getIconInfo cfg $ cfg) return) $ windowIds ws
+  -- XXX: Only one of the two things being zipped can be an infinite list, which
```
teleshoes: haha, there must be a better way..
code is a little too dense for me to come up with one quickly, tho, so i cant put my money where mouths are.

src/System/Taffybar/IconImages.hs
```diff
@@ -0,0 +1,88 @@
...
+pixBufFromColor imgSize (r,g,b,a) = do
+  let sampleBits = 8
+      hasAlpha = True
+      colorspace = Gtk.ColorspaceRgb
+  pixbuf <- Gtk.pixbufNew colorspace hasAlpha sampleBits imgSize imgSize
+  Gtk.pixbufFill pixbuf r g b a
+  return pixbuf
+
+-- | Convert a list of integer pixels to a bytestream with 4 channels.
+pixelsARGBToBytesRGBA :: [Int] -> [Word8]
```
teleshoes: btw, if youre looking for places to improve performance further, this is a good place.
i wrote this more to show exactly how it worked than to be performant.

src/System/Taffybar/WorkspaceHUD.hs
```diff
@@ -0,0 +1,504 @@
...
+    controllersList <-
+      mapM
+      (\(idx, controller) -> do
+         newController <- (updateController idx controller)
+         return (idx, newController)) $
+      M.toList controllers
+    return $ M.fromList controllersList
+
+updateWorkspaceControllers :: Context -> IO ()
+updateWorkspaceControllers c@Context { controllersVar = controllersRef
```
teleshoes: this is nice

teleshoes:
as for your kind offer, i would need (at least) the following to reproduce my WSS:
-`(String->String)` for: activeWS, hiddenWS, emptyWS, urgentWS (returns pango markup for label)
-gap between workspaces in pixels (i have it 3 on one machine, 5 on another)
-a colored border with the inside bg[NORMAL] the same as the taffybar bg[NORMAL]. i havent been able to figure out how to do this without using GtkFrame yet myself (the obvious soln is two containers with padding on the inside container and bg-color on the outside container, but then the inside container inherits the bg-color from the outside container, and you have to explicitly set a different one. with gtkframe, the inner hbox does not inherit the bgcolor)
-`customIcon` and `preferCustomIcon`. i see you have custom icons, but it looks different. in particular, i need to be able to give preference (or not) to the custom icon-overrides based on title/class
-`minIconCount`, to fill in the missing ones with blank icons that take up the same room.

im also quite concerned with your change re: icon updating. my current impl updates the icons FAR more often than necessary, but im afraid yours doesnt update often enough. it should ideally update the icons when any window title, class, or icon-hint changes.

IvanMalison:
> as for your kind offer, i would need (at least) the following to reproduce my WSS:
> -(String->String) for: activeWS, hiddenWS, emptyWS, urgentWS (returns pango markup for label)

Having a separate String->String for each type is silly. Instead it should be a single (Workspace->String)



> -customIcon and preferCustomIcon. i see you have custom icons, but it looks different. in particular, i need to be able to give preference (or not) to the custom icon-overrides based on title/class

it is different, but you can get the same behavior. 

See windowTitleClassIconGetter. The first argument will determine whether or not customIcons are preferred. You then set getIconInfo to the result of this value.

Having a getIconInfo is a much simpler, yet much more configurable way to allow the user to select the icon that they would like to have displayed for a given window.

teleshoes:
`Having a separate String->String for each type is silly. Instead it should be a single (Workspace->String)`
as long as workspact contains those states, then its equivalent in functionality. as far as ease and clarity go, its just a matter of having the complexity go in the user's `taffybar.hs` or in `WorkspaceHUD.hs`.
you think its silly because you wont use it, but its the default now and its quite tidy.

`See windowTitleClassIconGetter`
yep, upon closer inspection, looks perfect

`Why not just set the minWSWidgetSize option`
because then i have to calculate the widget size based on the current icon size, the frame border size, and the widget gap, all of which have separate configs, some of which depend on still more configs (like bar height, etc). this adds a ton of complexity to what is a simple boolean now, and introduces a dependency for further features (remember to update the min-widget-size when adding such-and-such)
also, styling the icon with its widget-name wont be possible if there is no icon.

`if you want a gtk frame all you have to do is add your own contents controller`
i had to re-do the border because of the problem you reported. it now uses a padded eventbox instead of a gtkframe, see my new pr.

its also important that the widget-names be set based on the current workspace-state so it can be custom-styled in the user's taffybar.rc

teleshoes:
```
-- | Set the widget names of the workspace button components:
-- border    => Workspace-Border-<WORKSPACE_NAME>-<WORKSPACE_STATE>
-- image     => Workspace-Image-<WORKSPACE_NAME>-<WORKSPACE_STATE>
-- container => Workspace-Container-<WORKSPACE_NAME>-<WORKSPACE_STATE>
-- label     => Workspace-Label-<WORKSPACE_NAME>-<WORKSPACE_STATE>
setWidgetNames :: Desktop -> String -> WorkspaceIdx -> IO ()
setWidgetNames desktop workspaceState wsIdx
  | Just ws <- getWS desktop wsIdx = do
      Gtk.widgetSetName (label ws)     (widgetName "Label"     (name ws))
      Gtk.widgetSetName (image ws)     (widgetName "Image"     (name ws))
      Gtk.widgetSetName (border ws)    (widgetName "Border"    (name ws))
      Gtk.widgetSetName (container ws) (widgetName "Container" (name ws))
  | otherwise = return ()
  where widgetName widget wsName = "Workspace"
                                   ++ "-" ++ widget
                                   ++ "-" ++ wsName
                                   ++ "-" ++ workspaceState
```

sakshamsharma:
Great work there! I like the underlines!
Although I agree with @teleshoes that fixed width would be quite the way to go for many people.

I couldn't build and test as of now, but I noticed that the mouse scrolling has not been implemented.
[Here](https://github.com/IvanMalison/taffybar/blob/dbf51aac25035ec9bdef50332850c13fcab5a090/src/System/Taffybar/WorkspaceSwitcher.hs#L237) is the relevant part from WSS. The relevant commit is 974fee733f0d76c301226741f1cb08c73a3fbd64 (as you can see, not a huge job).

sakshamsharma:
Okay. I just got to test this. Looks awesome! There are plenty of build warnings right now though. I suppose you're still in the process of resolving them.
Some points:
1. Is it possible to have an configuration to only display one window's icon per workspace? Perhaps the window which had the most recent focus.
2. What about hiding the workspace number if there's an icon already? I hate how the  workspace number is on the left and the icon is on the right. Things should be centered IMO :smile: So it may be nice to have this as an option.
3. The scrolling is still not there. Once you've stabilized your PR, I can add that perhaps.

IvanMalison:
> Okay. I just got to test this. Looks awesome! There are plenty of build warnings right now though. I suppose you're still in the process of resolving them.

What build warnings? There should only be 2, and I actually am thinking of leaving those.

> Some points:
> 
> Is it possible to have an configuration to only display one window's icon per workspace? Perhaps the window which had the most recent focus.

Yes, there is a minIcons (and a maxIcons) setting. If you set it to one you will get exactly this behavior.

> What about hiding the workspace number if there's an icon already? I hate how the workspace number is on the left and the icon is on the right. Things should be centered IMO  So it may be nice to have this as an option.

I like seeing the workspace number so I wouldn't want to make this the default, but the beauty of this pull request is that you can now write little peices of code to specify the exact layouts that you want without making changes to taffybar itself. If you want me to write something with this behavior for you, I'm more than happy to do so.

> The scrolling is still not there. Once you've stabilized your PR, I can add that perhaps.

Working on this right now

IvanMalison:
> Another factor: I think you should rather do an interactive rebase and clean up the history of this PR. The commit messages aren't really worthy of being put into master right now. I suppose a squash and merge for a huge PR isn't a good thing. That's for @travitch to decide though.

I generally like to avoid rewriting history. I also like having more granular commits because it allows you to undo more stuff and see more intermediate states.

sakshamsharma:
Well, you should probably at least move the WIPs into a single commit :smile: That would've been my personal choice, but I suppose it is, after all, an opinion.
There's another issue. Why do all workspaces show up for me? I remember this being an issue a while ago, I had fixed it [here](https://github.com/travitch/taffybar/commit/71f57f81f03128a3335f2e4186703b3bc3f98fdc). I just moved back to Xmonad from i3 (don't judge me, Xmonad had been broken for me since 2 months :stuck_out_tongue:  ), and I again see all my workspaces popping up. All empty workspaces are getting displayed as well.

IvanMalison:
@teleshoes I have added a function `hudFromPagerConfig` which takes your pagerConfig and makes an hud configuration that should provide nearly (except for improvements) identical behavior.

you can also build your taffyPager with taffyPagerHUDLegacy which uses this function.

> -(String->String) for: activeWS, hiddenWS, emptyWS, urgentWS (returns pango markup for label)

hudFromPagerConfig takes all of these and combines them into a single getIconInfo function.

> -gap between workspaces in pixels (i have it 3 on one machine, 5 on another)

This was already there

> -a colored border with the inside bg[NORMAL] the same as the taffybar bg[NORMAL]. i havent been able to figure out how to do this without using GtkFrame yet myself (the obvious soln is two containers with padding on the inside container and bg-color on the outside container, but then the inside container inherits the bg-color from the outside container, and you have to explicitly set a different one. with gtkframe, the inner hbox does not inherit the bgcolor)

I looked at #182 and reimplemented it in workspaceHUD. This sort of highlights the whole point of the new design. All of the added code can exist completely apart from any of the existing code. In theory, you could just add the below to your taffybar.hs to get the behavior you desire.

```
data WorkspaceBorderController =
  WorkspaceBorderController { border :: Gtk.EventBox
                            , borderContents :: Gtk.EventBox
                            , insideController :: WWC
                            }

buildBorderController :: ParentControllerConstructor
buildBorderController contentsBuilder context workspace = do
  cc <- contentsBuilder context workspace
  brd <- Gtk.eventBoxNew
  cnt <- Gtk.eventBoxNew
  Gtk.eventBoxSetVisibleWindow brd True
  Gtk.containerSetBorderWidth cnt (borderWidth $ hudConfig context)
  Gtk.containerAdd brd cnt
  Gtk.containerAdd cnt $ getWidget cc
  return $ WWC $ WorkspaceBorderController { border = brd
                                           , borderContents = cnt
                                           , insideController = cc
                                           }

instance WorkspaceWidgetController WorkspaceBorderController
  where
    getWidget bc = Gtk.toWidget $ border bc
    updateWidget bc wu@(WorkspaceUpdate workspace) =
      (Gtk.widgetSetName (border bc) $ getWidgetName workspace "Border") >>
      (Gtk.widgetSetName (borderContents bc) $ getWidgetName workspace "Container") >>
      (updateBorder bc wu)
```

> -customIcon and preferCustomIcon. i see you have custom icons, but it looks different. in particular, i need to be able to give preference (or not) to the custom icon-overrides based on title/class

This will also work exactly as it did with WSS

> -minIconCount, to fill in the missing ones with blank icons that take up the same room.
You'll get your desired behavior by setting min and max icons to 1

> 
> im also quite concerned with your change re: icon updating. my current impl updates the icons FAR more often than necessary, but im afraid yours doesnt update often enough. it should ideally update the icons when any window title, class, or icon-hint changes.

This is fixed. We now update whenver title, class or the icon itself changes.


I'm pretty sure that I have completely duplicated all of WorkspaceSwitchers functionality. You can consider any behavior that is different (and not unambigously just better such as update triggering) a bug. I really appreciate any help with testing.

IvanMalison:
@sakshamsharma the compile error is fixed

> There's another issue. Why do all workspaces show up for me? I remember this being an issue a while ago, I had fixed it here. I just moved back to Xmonad from i3 (don't judge me, Xmonad had been broken for me since 2 months  ), and I again see all my workspaces popping up. All empty workspaces are getting displayed as well.

workspaceHUD takes a different approach to this. You can hide or show the entire widget with the
`showWorkspaceFn :: Workspace -> Bool` 

the helpful value hideEmpty is also provided

you can use it like

``` haskell
defaultWorkspaceHUDConfig { underlineHeight = 3
                                            , showWorkspaceFn = hideEmpty
                                            }
```

With that said, when I use the legacy configuration setup, empty workspaces do not appear (unless images are turned on, in which case `const ""` is not enough.

teleshoes:
@IvanMalison wow, nice stuff man! the improvements look great. ill rebase on it as soon as i get a chance and let you know how it compares.

`I generally like to avoid rewriting history. I also like having more granular commits because it allows you to undo more stuff and see more intermediate states.`
yea! though, imo, you SHOULD squash typo commits or other trivial fixups; every commit should accomplish a specific, meaningful goal, whenever possible.
im a strong believer in rewriting history. it should only be taboo for changesets merged upstream or otherwise shared in a way that implies they wont change. imo, PRs definitely do not apply.

sakshamsharma:
Okay, the compile error is fixed. I can confirm the scrolling is working too. Once again, great job!! Some notes:

* When I implemented scrolling, I was a Haskell noob, and couldn't figure out how to exclude empty workspaces from popping up on scrolling. Can you do that? Travitch wanted that, afair.

* Weirdly enough, try keeping all your workspaces empty (with `hideEmpty` enabled). Move from 1 to 9 (assuming you have numeric named workspaces). You will notice that the workspace number shifts from the complete left (for `1`) slowly to the middle (`4` to `9`). Why is this happening? It looks very unsettling. The number should've stayed at the same place.

* Is there a way to limit the number of icons per workspace? (Not one I saw in the code). That'd be great, you know. And I fathom, not very difficult. An extra entry in the config, and a `take` clause while displaying icons, I suppose. Would help people who don't want sizes to become variably large.

* My bias against lots of commits is the same as that of @teleshoes . Typos etc can be merged with other commits. The rest is okay. I suppose this big a PR ought to have multiple commits.

sakshamsharma:
A bug report: https://ibb.co/eLWz6F
This shouldn't happen, right? I closed a terminal, and I'm stuck with this now. Even if I open a window in another other workspace now, nothing happens to taffybar. The only way something changes is when I open something in workspace 4 (where I closed that terminal). It looks normal that way. But nothing else seems to show up.

IvanMalison:
> A bug report: https://ibb.co/eLWz6F
> This shouldn't happen, right? I closed a terminal, and I'm stuck with this now. Even if I open a window in another other workspace now, nothing happens to taffybar. The only way something changes is when I open something in workspace 4 (where I closed that terminal). It looks normal that way. But nothing else seems to show up.

@sakshamsharma Can you send me the config that is causing this. Are you using hideEmpty?

IvanMalison:
> When I implemented scrolling, I was a Haskell noob, and couldn't figure out how to exclude empty workspaces from popping up on scrolling. Can you do that? Travitch wanted that, afair.

Yeah, that sounds easy enough.

> 
> Weirdly enough, try keeping all your workspaces empty (with hideEmpty enabled). Move from 1 to 9 (assuming you have numeric named workspaces). You will notice that the workspace number shifts from the complete left (for 1) slowly to the middle (4 to 9). Why is this happening? It looks very unsettling. The number should've stayed at the same place.
> 
> Is there a way to limit the number of icons per workspace? (Not one I saw in the code). That'd be great, you know. And I fathom, not very difficult. An extra entry in the config, and a take clause while displaying icons, I suppose. Would help people who don't want sizes to become variably large.

Yep, minIcons and maxIcons are both in WorkspaceHUDConfig. If you set them to the same value there will be no variable width.

> 
> My bias against lots of commits is the same as that of @teleshoes . Typos etc can be merged with other commits. The rest is okay. I suppose this big a PR ought to have multiple commits.

I think I'll fixup all the work in progress commits, but I like having the other ones, and I actually put a lot of thought into where I made commits.

sakshamsharma:
> @sakshamsharma Can you send me the config that is causing this. Are you using hideEmpty?

[Here](https://github.com/sakshamsharma/dotfiles-genie/blob/master/homefiles/taffybar/taffybar.hs) is my taffybar config. As you can see, it is almost completely your config, with a bit of my custom harmless quirks. I don't think it has any issue, except that, Yes, I use hideEmpty. I hope this helps in checking if the empty workspace slot was a reproducible bug.

> This sounds like super weird behavior, it sounds like there is a bug with hiding and showing workspaces, can you record a gif and send me your config?

As for the numbers shifting around, [here](https://giphy.com/gifs/l44QfHY5Tf0HYfPnW) is a GIF for that (apologies for the bad quality, I just ripped off how to recording GIF from stackoverflow :smile:

> I think I'll fixup all the work in progress commits, but I like having the other ones, and I actually put a lot of thought into where I made commits.

Sounds great!

IvanMalison:
> As for the numbers shifting around, here is a GIF for that (apologies for the bad quality, I just ripped off how to record GIF from stackoverflow 😄

Ah okay, I have some suspicions about what is happening. Can you set a minIcons value? Something like 3, and see if it continues to happen?

teleshoes:
wow, `hudFromPagerConfig` works perfectly for my setup. it looks identical and seems to function the same. ill use it for the next couple days and report any issues, but it looks great so far.

thanks for all your hard work!

i have one more request, though, before i can switch comfortably.
can you pull listening for `_NET_WM_DESKTOP` events out to a hud config option?
default-enabled is fine with me of course.

with it enabled, performance is horrible. in fact, i started writing about how WSHUD was worse than WSS before i remembered that while i reverted commit `1a5c422` in WSS, similar code was probably still in WSHUD. (removing line 269 makes performance more or less equivalent.)


**my typical test for icon performance is:**
-open 3 windows with different icons in one workspace
-cycle between them quickly (by holding alt+tab for 20s)
-observe cpu usage and subjective responsiveness of the ws icon

**WITHOUT `_NET_WM_DESKTOP`**
    ```
    the icons flash by as quickly as the window focus travels
    ```

**WITH `_NET_WM_DESKTOP`**
    ```
    thc icon changes once and then the entire taffybar hangs frozen until i release alt+tab. it remains frozen for another 3-10s after i stop, and THEN all of taffybar un-hangs, my clock updates, and the icon updates. if i press alt+tab again without waiting the full 3s, it never un-hangs.
    ```

teleshoes:
i see weird scroll order or something as well, btw. i DO use rxvt, but i have a customIcon for it. it seems to go `1->2->3->4->5->4->5->6->7->6->7->8->9->1...`

IvanMalison:
> i see weird scroll order or something as well, btw. i DO use rxvt, but i have a customIcon for it. it seems to go 1->2->3->4->5->4->5->6->7->6->7->8->9->1...
> 
> EDIT: i dont hide-empty, so i just see the focus moving from 4 to 5 back to 4 and then back to 5. it seems to happen when moving from an empty workspace to a workspace that has urxvt, or something like that.

I'm not really sure what would account for this because the scrolling code is not really appreciably different from what is in WSS.

> i have one more request, though, before i can switch comfortably.
> can you pull listening for _NET_WM_DESKTOP events out to a hud config option?
> default-enabled is fine with me of course.

Yeah, that seems like a reasonable option. Note that if a window updates its icon without updating its tile or window class, the icon in the workspaceHUD will not update.


> 
> with it enabled, performance is horrible. in fact, i started writing about how WSHUD was worse than WSS before i remembered that while i reverted commit 1a5c422 in WSS, similar code was probably still in WSHUD. (removing line 269 makes performance more or less equivalent.)
> 
> my typical test for icon performance is:
> -open 3 windows with different icons in one workspace
> -cycle between them quickly (by holding alt+tab for 20s)
> -observe cpu usage and subjective responsiveness of the ws icon
> 
> WITHOUT _NET_WM_DESKTOP
> the icons flash by as quickly as the window focus travels
> 
> WITH _NET_WM_DESKTOP
> thc icon changes once and then the entire taffybar hangs frozen until i release alt+tab. it remains frozen for another 3-10s after i stop, and THEN all of taffybar un-hangs, my clock updates, and the icon updates. if i press alt+tab again without waiting the full 3s, it never un-hangs.


I made the updates to the icons as efficient as they can possibly be. Its interesting that you are having such severe issues because workspacehud is noticeably faster on my machine.

Can you try logging the calls into the onIconUpdated callback. For me they happen relatively infrequently and don't really have a huge effect.

IvanMalison:
> can you pull listening for _NET_WM_DESKTOP events out to a hud config option?

Done. I also got rid of a all of the WIP commits from the early development phase, I'd like to leave the remaining ones in though. The commit messages are all pretty clear, and if you look at each commit it makes one logical change.

IvanMalison:
@travitch I think that this is ready to get merged. Let me know if you have any objections (or if you would be willing to consider having workspaceHUD completely replace workspaceSwitcher).

teleshoes:
sorry to say that so far, taffybar has crashed twice since i switched. i think both times it involved urgent, but im not sure. (i lost the output because my launcher script stupidly clobbered the log. this has been fixed so if it happens again, ill have the output(
also, i have not yet rebased on the latest, and i had edited out the net-wm-desktop event.

could be coincidence. probably isnt.

sakshamsharma:
@IvanMalison 
> Ah okay, I have some suspicions about what is happening. Can you set a minIcons value? Something like 3, and see if it continues to happen?

I set both min and max to 1, and the issue does not appear any more. The numbers stay aligned to the left as expected.

> Also do you have a terminal like urxvt that does not set an EWMH icon active in one of those workspaces?

Well, I use urxvt. But none of those workspaces had any window open, not even urxvt.

Another note, isn't it confusing to have `minIcons :: Int` and `maxIcons :: Just Int` ? Just saying :smile: There *might* be a better way, but I can't think of it. Let's not dwell on this for now though. This PR is huge enough already.

IvanMalison:
@teleshoes I get crashes from #105 from time to time. I haven't really seen anything else. Hopefully your logs give us something helpful.

teleshoes:
@IvanMalison the crash is reproducible and is caused by WSHUD

if i open up pidgin, have two conversations in one window, and switch between them rapidly with ctrl+tab, WSHUD crashes the entire taffybar within 5s or so. WSS does not.

also, WSHUD occasionally crashes while switching conversations normally (without doing it rapidly).
seems like maybe its an edge case.

usually, taffybar just permanently freezes. (ive waited for ~5min for it to unfreeze and it doesnt.) however, sometimes it crashes with an error:
`Gdk:ERROR:/build/gtk+2.0-1aCJs4/gtk+2.0-2.24.31/gdk/gdkregion-generic.c
Aborted`

note that pidgin sets the _NET_WM_ICON based on the current buddy's icon, so the icon is rapidly changing. this works fine in WSS.

also note that in BOTH testing scenarios, i am not using _NET_WM_DESKTOP.

IvanMalison:
> @IvanMalison the crash is reproducible and is caused by WSHUD
> 
> if i open up pidgin, have two conversations in one window, and switch between them rapidly with ctrl+tab, WSHUD crashes the entire taffybar within 5s or so. WSS does not.

I can't reproduce a crash, but I can get taffybar to momentarily hang if I hold down ctrl+tab long enough.

The reason for the discrepancy in behavior is that workspaceHUD CORRECTLY subscribes to
"_NET_WM_ICON", while workspace switcher does not, the only reason that workspace switcher is SOMETIMES able to handle icon changes is because the title/class of the window often changes in conjunction with the icon. When this is not the case, the icon WILL fail to update with WSS.

There is an existing option updateOnWMIconChange in workspaceHUD config that allows the user to disable this callback if they please which should make the behavior of workspaceHUD identical to that of workspace switcher.

To be clear, I do not think that @teleshoes bug is in fact a bug at all, it is an improvement in behavior that some might want to disable.

IvanMalison:
@teleshoes Wow this is very strange. Though I am not able to reproduce your bug with WorkspaceHUD, I AM able to reproduce it with WSS.

Are you sure you didn't get the two mixed up somehow. I'm going to do a screenrecording so I can show the crash happening with WSS, and then failing to happen with WSHUD.

teleshoes:
haha, quite sure!
this is now happening to me about once per day, and never did before. i was also quite careful in my tests. one test called `wspaceSwitcherNew`, and one called `buildWorkspaceHUD`, and i repeated both multiple times.

also, i would agree with you on your assessment of its bug-vs-feature-ness, EXCEPT that taffybar somtimes **crashes and dies** and needs to be restarted. this i can never call "an improvement in behaviour".

teleshoes:
just reproduced crash+die handily in WSHUD after failing for 5 min to get it in WSS:
`Gdk:ERROR:/build/gtk+2.0-1aCJs4/gtk+2.0-2.24.31/gdk/gdkregion-generic.c:1110:miUnionNonO: assertion failed: (y1 < y2)`

teleshoes:
ALSO, ive been failing to mention i get warnings....sorry, quite possibly very relevant. i do not get these warnings in WSS, and i assume that they are from using the border feature.
```
(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed

(taffybar-linux-x86_64:2975): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed
```

teleshoes:
screenshot-ed the difference between my test cases. first one (WSHUD) crashes reliably within 5s of ctrl+tab, usually with a GTK assertion, sometimes with memory-corruption, second one ive never seen crash.

IvanMalison:
> also, i would agree with you on your assessment of its bug-vs-feature-ness, EXCEPT that taffybar somtimes crashes and dies and needs to be restarted. this i can never call "an improvement in behaviour".

I can not personally reproduce this despite trying for quite some time. Here is workspaceHUD CONSISTENTLY recovering from the pidgin experiment: https://gfycat.com/UnrealisticRewardingBengaltiger

And here is workspaceswitcher hanging with the same test

https://gfycat.com/FaroffImmaculateAfricanmolesnake

sakshamsharma:
So I came across that bug again, where the total workspaces displayed gets trimmed at 3/4. It has been happening very frequently for me (once every hour or so, when I'm not switching workspaces too often). Have not managed to reproduce it reliably, but here's the description:

I have the following workspaces:
1 2 3 4 5 6 7 8 9 IM Mail NSP

Out of these, the following were open since a while (I hadn't moved away from these in a while):
1 2 3 Mail
Each of them had a window or 2 open (none were URxvt, I have Emacs for that :smile:)

I moved to workspace 4 for some job, and noticed that WS3 disappeared from the HUD. I moved back to WS3, and it came back. To try a bit more, I opened a new window in WS4, and then moved to WS5. Voila, both WS3 and WS4 disappeared, and 5 came up. At this point, the display had the following:
1 2 5 Mail

I played around a bit more. On moving to WS7, it replaced WS2, and the display became:
1 7 3 Mail

The same was true for WS8. It seemed as if something had caused the internal state to become inconsistent.

One thing to note is that I do NOT use Taffybar's PagerHints. Here's the relevant line from my Xmonad config:
https://github.com/sakshamsharma/xmonadic-zest/blob/master/xmonad.hs#L50

In the end, I just restarted Taffybar. Not a big deal, but just letting you know that there's an issue still lurking (assuming that you don't rely on the pagerHints for something).

IvanMalison:
> One thing to note is that I do NOT use Taffybar's PagerHints.

The only thing that taffybar's pagerhints do is expose a property that allows the pager to see which workspaces are visible (but not necessarily active). If you only have one screen it should have no effect.



> I moved to workspace 4 for some job, and noticed that WS3 disappeared from the HUD. I moved back to WS3, and it came back. To try a bit more, I opened a new window in WS4, and then moved to WS5. Voila, both WS3 and WS4 disappeared, and 5 came up. At this point, the display had the following:

Just to be clear:
Did you try going back to workspace 4 to make sure that a window was still there?

> I played around a bit more. On moving to WS7, it replaced WS2, and the display became:
> 1 7 3 Mail

This is really strange because the workspaces are out of order. The workspaces are never actually removed from the original widget (just hidden and unhidden) So it seems REALLY strange that something like this would happen.

> The same was true for WS8. It seemed as if something had caused the internal state to become inconsistent.

The thing thats weird about this is that workspaceHUD actually regenerates the ENTIRE state of all of the workspaces every time an update is triggered, so it doesn't really make sense that the state would be inconsistent because, well there isn't much state to speak of.

The only reason any state is really kept at all is for the sake of avoiding redrawing images that don't need to be redrawn.

Unfortunately, I'm sort of at a loss about how to help you here. The best thing I can think of is to try to log hud's perspective on the workspace state at every redraw.

Obviously, a reliable repro would be super helpful!

sakshamsharma:
> Did you try going back to workspace 4 to make sure that a window was still there?

Yes I did. It was there.

>  The best thing I can think of is to try to log hud's perspective on the workspace state at every redraw.

Running busy, but will try to do that soon. Or give a reliable repro if I can find one. Thanks!

simlu:
My work around for any crashes is just to automatically restart. See #105. Problem solved :)

teleshoes:
heh, thats hardly 'problem solved', though it is making the best of a bad situation.

as it is, the crashing in WSHUD so annoying that ill be sticking with WSS for now. sorry, @IvanMalison, this is really good work, and i look forward to eventually using it.

IvanMalison:
@teleshoes I've had my taffybar running for 72 hours straight now.

How often is your taffybar actually crashing? Is it reproducible? Did you see my video above doing the exact thing that you claimed causes a crash?

teleshoes:
UGGGHHHHH i was not using the latest WSHUD....
you snuck an update in while i wasnt looking. ;)
anyway, ill switch back to WSHUD for a few more days and hope the problems go away.

wshud HAS been crashing 2-3x a day at random times.
my pidgin test reliably breaks it (old version still), but it would crash at times when there was nothing weird going on, as far as i could tell. ill let you know if the updates affect it.

src/System/Taffybar/WorkspaceHUD.hs
```diff
@@ -0,0 +1,728 @@
...
+                        , hudConfig = cfg
+                        , hudPager = pager
+                        }
+
+  -- This will actually create all the widgets
+  updateAllWorkspaceWidgets context
+
+  mapM_ (subscribe pager (onActiveChanged context))
+        [ "_NET_CURRENT_DESKTOP"
+        , "_NET_WM_DESKTOP"
```
teleshoes: THIS is the problematic event, not _NET_WM_ICON...

teleshoes:
updated to latest.
unfortunately, crash is still easily reproducible in my pidgin experiment in WSHUD, and not at all in WSS.
(i have again removed manaully from your code _NET_WM_DESKTOP, as both WSS and WSHUD slow to a crawl with it on my system. i have no problem with _NET_WM_ICON)

i will let you know if it crashes by chance tomorrow.

teleshoes:
still get these warnings from border, btw: `(taffybar-linux-x86_64:6463): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed`

sakshamsharma:
A weird screenshot :stuck_out_tongue: 


`NSP` is for my scratchpad. Also, if I moved back to a sane workspace like `3`, and then moved to `4`, it would overwrite the `Mail` workspace. Just in case the weird location of `NSP` helps you figure out something.

IvanMalison:
@sakshamsharma Yeah, I was finally able to replicate this issue. It's so weird, because those widgets are never added or removed. I have an idea for something that might potentially mitigate it, but this makes so little sense to me. It could be a weird gtk bug...

IvanMalison:
@sakshamsharma Do you do a bunch of display switching?

i.e. do you plug in and then unplug monitors to your computer?

sakshamsharma:
No. No display switching at all.

IvanMalison:
@sakshamsharma I think I may have found the issue. There was an extra call to Gtk.containerAdd.

It turns out that this was also what was causing the output that @teleshoes mentioned

```
(taffybar-linux-x86_64:18553): Gtk-CRITICAL **: gtk_box_pack: assertion 'child->parent == NULL' failed
```

I can see how this might cause the issue that you were experiencing @sakshamsharma basically because the widget is added to the container twice, it gets confused in certain situations.

sakshamsharma:
Great! Thanks a lot ! I'll let you know if it occurs again. Else, assume it isn't :smile:
Travitch must be busy, he hasn't reviewed #164 as well. I have to stash and apply the changes everytime I want to build your code.

Oh forgot to mention, my tests are on GHC 8.

IvanMalison:
@travitch Do you need help with taffybar maintenence? I think you should consider adding some contributors to the project who can help out.

IvanMalison:
@sakshamsharma Hmmm I think maybe we just want to use containerAdd instead of boxPack

IvanMalison:
Yeah, make sure to use the newest one I put up there.

sakshamsharma:
Yeah I `reset --hard`ed to your remote branch.
@IvanMalison I think you meant adding more maintainers?

Also, I think you were wrong when you said pagerHints aren't used. Without pager hints, the current layout does not get displayed in the HUD. I finally managed to recompile xmonad with taffybar's pagerHints using the following:
```
stack build && stack install taffybar; stack exec taffybar& # Inside taffybar source
stack exec xmonad -- --recompile
stack exec xmonad -- --restart
```

IvanMalison:
> Also, I think you were wrong when you said pagerHints aren't used. Without pager hints, the current layout does not get displayed in the HUD. I finally managed to recompile xmonad with taffybar's pagerHints using the following:

I never said they weren't used, I just said that the only thing they influenced in the HUD itself is the display of visible workspaces, which is correct. You are also correct that it is used to show the current layout.

you can see what it does right here https://github.com/travitch/taffybar/blob/cbbbbb23af38b19e8a2f9406264277279b9a6d0d/src/System/Taffybar/Hooks/PagerHints.hs#L73

pretty obvious that it sets the layout and the visible workspaces.


> taffybar's pagerHints using the following:

Not sure why that's necessary. I've been able to compile with pagerHints for a while by simply adding taffybar as a dependency in stack.yaml

sakshamsharma:
Bug still persists :cry:
Bug still persists :/

teleshoes:
my tragic crashing bug still happens.

crashed twice today. all of taffybar just froze, and then stopped repainting gtk windows, no stack trace, no info to debug. there was no obvious trigger. maybe switching workspaces, maybe switching a window, maybe it just died at rest. either way, there was no apparent stress on taffybar or WSHUD.

back to WSS for me....

:(

IvanMalison:
> Also, what about these build warnings? I hope they are innocent and you know about them? They seem really easy to resolve IMO.

The first one is to support old versions of GHC where `<$>` is not imported by default.

The other two are cases where if you look at the code its clear that the other possibilites are simply impossible.

travitch:
To silence the Applicative warning, you can put an `import Prelude` after the import of `Control.Applicative`.  You'll probably see that in a few other taffybar modules for the same reason. It takes advantage of the way GHC searches for names: it will get the `<$>` name from Control.Applicative (first), rather than from Prelude.  If you don't explicitly import Prelude, it is implicitly the first import.  Note that this only works if you import Control.Applicative unqualified.

IvanMalison:
@sakshamsharma I've added a few things that might possibly fix the issue you've been seeing. I haven't seen it for a little bit so I'm hopeful that it might be fixed.

I've also added a debug mode that allows you to turn on logging and rebuild the bar on command. If you can enable that just in case it happens again to see if either is helpful for figuring out what is going on I would appreciate it.

@teleshoes I've added a rate limiter to all of the events that are generated
You claim that workspaceHUD crashes when you cycle with pidgin, but as I mentioned before, I find the exact opposite, especially now that workspaceHUD has rate limiting -- here is video evidence:

https://gfycat.com/@ivanmalison/detail/LastFloweryConey

I'd really like it if you could record a video to substantiate your claim that this causes crashes for HUD but not switcher, since again, I find the EXACT OPPOSITE in all of my tests.

sakshamsharma:
I have a comment. The current rate limit is what is called debounce in angular2 language. It waits for events for 100 milliseconds, and then executes. What would be much better is to have the following:

* Execute the first event
* If another event comes within 100ms of that event, don't update the HUD.
* Once 100ms have elapsed since the last event which was executed, see if you had a waiting event.
* If so, update the HUD again. Else, don't.

The current 100 milliseconds is quite annoying to look at, frankly. Although I see your view point, what I'm suggesting is much more code I think, and requires a mutable state in the most trivial implementation.

Also, even though the HUD does not update more frequently than 100ms, the window icon still changes without any rate limit.

teleshoes:
@IvanMalison heh. i DID watch your video, and i agree that you ARE in fact seeing the OPPOSITE of what im seeing.
did you see my screenshots of the crash? if you look on the left, you will see the code where HUD is enabled with taffybar crash on screen, and the code where WSS is enabled on the other, with no crash.

i will attempt a video recording to satisfy your need for moving-image-proof. ;)

teleshoes:
here is the video, link expires in 48h
https://expirebox.com/download/c625b50356c7c17e2c611cf11bab0e84.html

first, i uncomment WSHUD and compile. then i pidgin-test. then i uncomment WSS and compile and pidgin-test.

in WSHUD, you will see taffybar freeze, clock and all, but not die. (see the screenshots above for what the death looks like, after 5-15s of pidgin). the icon doesnt change at all.

in WSS, you will see the icon change quickly while clock is updating. (framerate of video doesnt do justice to how quickly the icon updates, but its still fast in the video)

IvanMalison:
> EDIT: tests are against commit 3903e32 Add an option to control whether updates due to "_NET_WM_ICON" occur, with line 271 _NET_WM_DESKTOP removed.

@teleshoes BTW - I added an option (updateEvents) to HUDConfig that allows you to specify which updates you would like the widget to be updated on. You can simply use the value

``` haskell
[ "_NET_CURRENT_DESKTOP"
, "_NET_DESKTOP_NAMES"
, "_NET_NUMBER_OF_DESKTOPS"
, "WM_HINTS"
]
```

IvanMalison:
> I have a comment. The current rate limit is what is called debounce in angular2 language. It waits for events for 100 milliseconds, and then executes. What would be much better is to have the following:
> 
> Execute the first event
> If another event comes within 100ms of that event, don't update the HUD.
> Once 100ms have elapsed since the last event which was executed, see if you had a waiting event.
> If so, update the HUD again. Else, don't.
> The current 100 milliseconds is quite annoying to look at, frankly. Although I see your view point, what I'm suggesting is much more code I think, and requires a mutable state in the most trivial implementation.
> 
> Also, even though the HUD does not update more frequently than 100ms, the window icon still changes without any rate limit.

@sakshamsharma Yeah, it's not the best rate limiting approach for this use case. I actually can't notice it at all, but it was the the only impl I could find on short notice. I might try writing my own now that is better suited to the needs of WorkspaceHUD.

teleshoes:
@IvanMalison have to ask; did you see my video, and are you quite convinced of my sanity now? :D

IvanMalison:
> @IvanMalison have to ask; did you see my video, and are you quite convinced of my sanity now? :D

haha yes, of course. I'm not really sure what to chalk it up to. Maybe different versions of gtk or X11? Are you also using XMonad? Its hard for me to try to fix something that I can't reproduce. I guess I could start by using your exact configuration of taffybar.

IvanMalison:
> have a comment. The current rate limit is what is called debounce in angular2 language. It waits for events for 100 milliseconds, and then executes. What would be much better is to have the following:
> 
> Execute the first event
> If another event comes within 100ms of that event, don't update the HUD.
> Once 100ms have elapsed since the last event which was executed, see if you had a waiting event.
> If so, update the HUD again. Else, don't.
> The current 100 milliseconds is quite annoying to look at, frankly. Although I see your view point, what I'm suggesting is much more code I think, and requires a mutable state in the most trivial implementation.
> 
> Also, even though the HUD does not update more frequently than 100ms, the window icon still changes without any rate limit.

@sakshamsharma Actually, it turned out that the rate limiting algorithm was essentially a leaky bucket algorithm, but since I was not firing the request on a separate thread, the request was waiting on a response from the worker thread, even though it was not going to be used. 996639a fixes this and everything should fire off pretty much instantaneously now unless there are way too many events.

IvanMalison:
@travitch ping. Any chance you can merge this soon?

travitch:
I like the premise, but haven't been following the details.  What is the consensus so far on stability?

IvanMalison:
@travitch It shouldn't affect anyone who doesn't opt in to using workspaceHUD it in any way, since it leaves the existing workspaceSwitcher code in place. I've been using this for about a week now with no issues, but @teleshoes (I'm not sure if he's tried the latest version) seems to have an issue that I have not been able to replicate. I started getting @sakshamsharma's bug when I used a different machine, but that issue has since been fixed AFAICT.

WHUD could theoretically completely replace WSS in the future, but I imagine that @teleshoes might object to this until his issue is completely resolved. Since this pull request has already gotten rather large, and the probability of impact on WSS users is quite small, I think its probably better to just merge this now, and deal @teleshoes bug in a separate PR.

PR #181, opened 2017-02-18
Made NetMonitor capable of handling multiple interfaces

eskimor:
By summing the speeds of all interfaces actually available. It also
handles interfaces well which don't exist at all (e.g. a USB network
card). By simply summing the speeds of all available devices you
always*) get a valid speed information regardless of which network
interface is currently in use/available.

*) False values will be shown once every time a device appears or vanishes.

I just hacked this together - it seems to work for me!

### DIFF
--- a/src/System/Information/Network.hs
+++ b/src/System/Information/Network.hs
@@ -17,26 +17,27 @@
 module System.Information.Network ( getNetInfo ) where
 
 import Control.Applicative
+import Control.Monad
+import Control.Exception (catch, SomeException)
 import Data.Maybe ( mapMaybe )
 import Safe ( atMay, initSafe, readDef )
 import System.Information.StreamInfo ( getParsedInfo )
+import Control.Monad.Trans.Maybe (MaybeT(..))
 
 import Prelude
 
 -- | Returns a two-element list containing the current number of bytes
 -- received and transmitted via the given network interface (e.g. \"wlan0\"),
 -- according to the contents of the @\/proc\/dev\/net@ file.
-getNetInfo :: String -> IO (Maybe [Integer])
-getNetInfo iface = do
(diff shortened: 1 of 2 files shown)


eskimor:
Do you really still require patches to work with ghc 7.8? If so I will obviously try to make it work - I assume traverse was differently defined back then (only on lists).

eskimor:
Use case: laptop with WIFI, at work you have a docking station when on the road you use your mobile's internet connection via bluetooth: Simply list all interfaces you might have and you always get speed values in your taffybar, e.g. for me I have:

```
netMon = netMonitorMultiNew 2.0 ["wlp58s0", "bnep0", "enp5s0u1u2", "enp57s0u1u2"]
```

by introducing

```
netMonitorMultiNew
netMonitorMultiNewWith
```
this PR maintains backwards compatibility with the old behaviour.

eskimor:
woohooo - cool, thanks for merging! :-)

travitch:
I did some cleanup in the netmonitor code to remove some partiality - could you give it a whirl to make sure it still works for you?  The relevant commit is edea6c650fb0e6c1f609492d7ac4f956fecc095b

eskimor:
@travitch seems to work just fine! Thanks!

sakshamsharma:
Pointing out. I had the same functionality working this way earlier:
```
let  wnet = netMonitorNewWith 10 "wlp8s0" 1 myNetFormat
      enet = netMonitorNewWith 10 "enp9s0" 1 myNetFormat
      hud = taffyPagerHUDNew myPagerConfig hudConfig
      .......

  defaultTaffybar defaultTaffybarConfig { startWidgets = [ hud ]
                                        , endWidgets = [ wnet
                                                                 , enet
                                                                 , ......
```

PR #182, opened 2017-02-18
bugfix: GtkFrame => EventBox

teleshoes:
this is a simple bugfix to replace the border widget with an ebox instead of a frame.

the frame wont draw a background on certain versions of gtk, so the styling didnt work.
(thx @IvanMalison for reporting)

theres some small cleanup as well, but nothing user facing.
(users that dont use borders will see no change, users with borders that specify additional widget-styling will see the bug fixed)


src/System/Taffybar/WorkspaceSwitcher.hs
```diff
@@ -226,31 +231,34 @@ addButton :: Gtk.BoxClass self
...
-    ebox <- Gtk.eventBoxNew
-    Gtk.widgetSetName ebox $ name ws
-    Gtk.eventBoxSetVisibleWindow ebox False
-    _ <- Gtk.on ebox Gtk.buttonPressEvent $ switch idx
-    _ <- Gtk.on ebox Gtk.scrollEvent $ do
+    let brd = border ws
+    let con = container ws
+    btnParentEbox <- Gtk.eventBoxNew
+    iconLabelBox <- Gtk.hBoxNew False 0
+
```
teleshoes: pull all the widget-adding to one spot for clarity

taffybar.rc
```diff
@@ -40,8 +40,15 @@ style "taffybar-workspace-border-urgent" = "taffybar-default" {
...
+widget "*Workspace-Border-*-visible*"    style "taffybar-workspace-border-visible"
+widget "*Workspace-Border-*-empty*"      style "taffybar-workspace-border-empty"
+widget "*Workspace-Border-*-hidden*"     style "taffybar-workspace-border-hidden"
+widget "*Workspace-Border-*-urgent*"     style "taffybar-workspace-border-urgent"
+
+widget "*Workspace-Container-*-active*"  style "taffybar-default"
+widget "*Workspace-Container-*-visible*" style "taffybar-default"
+widget "*Workspace-Container-*-empty*"   style "taffybar-default"
+widget "*Workspace-Container-*-hidden*"  style "taffybar-default"
+widget "*Workspace-Container-*-urgent*"  style "taffybar-default"
```
teleshoes: necessary so that the inside of the button is styled as the rest of taffybar, instead of like the border.

Issue #139, opened 2015-12-09
Running multiple instances


Tordek:
I run a dual head setup, and Taffybar's widgets would be a great replacement for my setup. However, since Taffybar requires the config file to be in a particular location, it's not possible (except with some bash script, I suppose) to run two instances, each with a different config.

Either that, or a second options would be to be able to have more than one Taffybar window on the same config file.

The first option seems simpler, though.

sakshamsharma:
I used to keep a sed script to launch 2 instances.
So it would first start one instance of taffybar. Then it would use the ```sed``` command to replace the monitor number in the config file. Relaunches taffybar. Then uses sed again to restore the file to its original state.
Hacky workaround, but works.

alem0lars:
can you share the script ?

IvanMalison:
@alem0lars See What I wrote in #61 for a better approach

IvanMalison:
This should probably be closed now

Issue #61, opened 2014-01-11
Multimonitor support


mathstuf:
I'm the developer/maintainer of X.H.DynamicBars in xmonad-contrib. It'd be nice if taffybar integrated with it. I see a few ways it could be done
- DBus method call for monitor add/remove events (if keeping a single taffybar instance which draws multiple bars)
- Take a monitor number on the command line (multiple instances; one per screen)

Maybe there's a better way?

mathstuf:
I don't see a commit that does that. Is it pushed?

IvanMalison:
@mathstuf it probably isn't in taffybar -- you can just do it in the main of your config
``` haskell
main = do
  monString <- getEnv "TAFFYBAR_MONITOR"
  let monitorNumber = fromMaybe 1 $ readMaybe monString
  ... rest of your config, making sure to set
defaultTaffybar
    defaultTaffybarConfig
    { startWidgets = ...
    , endWidgets = 
    , monitorNumber = monitorNumber
      }
```

that will need imports
``` haskell
import Data.Maybe
import Text.Read
import System.Environment
```

alem0lars:
What if I detach the screen and reattach?

IvanMalison:
Just trigger a restart for each monitor that is active (you can use xrandr or whatever).

Alternatively, #183 will be amended shortly to handle this!

IvanMalison:
This can be closed now.

PR #183, opened 2017-03-04
Add support for multiple monitors

### LINKED ISSUE #153: One taffybar on each monitor?
Can taffybar be configured so that there is one instance spawned for each monitor? I imagine it would need some cooperation with xmonad or another supervisor which would ensure that instances are spawned and killed when monitors are dynamically added or removed. If I have skipped something obvious in the manual, sorry!

BTW, in TODO you write:

> Better behavior when adding/removing monitors (never tried it)

, but, apart from the above question, taffybar's support for adding/removing monitors seems nice to me. When I reconfigure xrandr, taffybar automatically moves to the new primary primary monitor and resizes itself so it will fit.

IvanMalison:
Fixes #153, #183, #61, #139

You'll need to set 
``` haskell
monitorFilter = allMonitors
```

To try it out.

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE TupleSections #-}
 -- | The main module of Taffybar
 module System.Taffybar (
   -- * Detail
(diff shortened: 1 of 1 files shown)


IvanMalison:
There is a performance penalty to running this, which makes me think that it may be better to have a separate process for each bar, and provide a script and a sample config that gets a value from the environment to set the monitor number.

This is certainly more straightforward though.

sakshamsharma:
What may be nicer is if it were possible to have Taffybar stretch across both monitors, with only selective information on each of them. But, small steps :smile:

BTW how big is the performance hit? I don't have 2 monitors right now to test, but I do run a multihead often.

IvanMalison:
@sakshamsharma yeah... that would be easy enough. The config would be a

monitorNumber -> Config type deal

PR #186, opened 2017-04-06
Per monitor toggle

### LINKED ISSUE #153: One taffybar on each monitor?

IvanMalison:
This exposes a mechanism by which taffybar can be toggled on and off. The toggling works on a per monitor basis (taffybar can be separately turned on and off on each monitor). The toggling mechanism is general in that it simply provides an IO action that refreshes the config of each taffybar. A webserver that listens for toggle requests is provided as the default way to handle toggling but other schemes could use the same config variables that are created here.

Fixes #153, #183, #61, #139


src/System/Taffybar.hs
```diff
@@ -175,12 +182,15 @@ data TaffybarConfig =
                  , endWidgets :: [IO Widget] -- ^ Widgets that are packed from right-to-left in the bar
                  }
 
+type TaffybarConfigEQ = (TaffybarConfig, StableName TaffybarConfig)
```
travitch: I assume you are using the `StableName` for equality tests later.  Is it possible to just do comparisons on `(screenNumber, monitorNumber)` instead?  Or is that not unique enough?  I can imagine someone might want to put two bars on one screen, but is that even possible right now?  Would mixing in the `barPosition` field be enough to get a unique key without requiring stable names?

src/System/Taffybar/ToggleMonitor.hs
```diff
@@ -0,0 +1,73 @@
...
+import           Control.Monad
+import           Control.Monad.Trans
+import           Control.Monad.Trans.Maybe
+import qualified Data.Map as M
+import           Data.Maybe
+import qualified Graphics.UI.Gtk as Gtk
+import           Graphics.UI.Gtk.Gdk.Screen
+import           System.Taffybar
+import           Text.Read hiding (get, lift)
+import           Web.Scotty
```
travitch: I have to say I'm a bit reluctant to bring an entire web server into the process.  Is there a reason to prefer something that heavy-weight?  We already have a d-bus dependency, which could serve much the same purpose.  A file interface might be even simpler.

src/System/Taffybar/ToggleMonitor.hs
```diff
@@ -0,0 +1,73 @@
...
+import           XMonad.Core ( whenJust )
+
+toggleableMonitors :: MV.MVar (M.Map Int Bool)
```
travitch: The type `M.Map Int Bool` stands out here - is it encoding more information than an `IntSet`?  If so, I think I'd prefer a more expressive type to make it clear that there are three states (`True`, `False`, and "not present").

src/System/Taffybar.hs
IvanMalison: The reason for using stablenames is so that it is possible to change the taffybar config dynamically for a particular bar at any time. You don't want to completely recreate the bar if nothing has changed, and normally you could just use (==) for this, but because taffybar has fields that are functions an IO operations, this is not possible.

There are definitely other ways to handle this, but this felt like the easiest way to handle this to me.

src/System/Taffybar/ToggleMonitor.hs
IvanMalison: Having a M.Map Int Bool just makes it easier (or more natural) to change the 'default value' when the bar hasn't yet been toggled on a particular monitor.

It also makes the code that toggles have fewer branches since you don't have to worry about whether you are doing an insert/delete operation, you just negate whatever the current value is.

``` haskell
toggleTaffyOnMon fn mon = do
        MV.modifyMVar_ enabledVar $ \numToEnabled -> do
          let current = fromMaybe True $ M.lookup mon numToEnabled
          return $ M.insert mon (fn current) numToEnabled
```

vs

``` haskell
toggleTaffyOnMon fn mon = do
        MV.modifyMVar_ enabledVar $ \enabledMonitors -> do
          let current = elem mon enabledMonitors
          return $ if current
                   then delete mon enabledMonitors
                   else insert mon enabledMonitors
```

All that said, I really don't have strong feelings about this and I'd be happy to change it in the dbus version that I'm about to write if you would prefer

src/System/Taffybar/ToggleMonitor.hs
IvanMalison: Yeah dbus didn't even occur to me but would almost certainly be better.

src/System/Taffybar.hs
IvanMalison: > I assume you are using the StableName for equality tests later. Is it possible to just do comparisons on (screenNumber, monitorNumber) instead?

No, because the TaffybarConfiguration that is being provided for the monitor could also change. This is an important feature to me, at least for the time being, because the tray widget can only be placed on 1 monitor, and I need to have the ability to move this around (when I plug in/unplug my laptop).

src/System/Taffybar/ToggleMonitor.hs
IvanMalison: Okay I replaced scotty with dbus. Let me know if you'd like different names for the
ObjectPath, Interface or connection.

IvanMalison: dbus can be invoked using
```
dbus-send --print-reply=literal --dest=taffybar.toggle /taffybar/toggle taffybar.toggle.showOnMonitor int32:0
```

IvanMalison:
Hmm the dbus version seems to befailing with

```
Apr 08 03:27:57 imalison-arch env[2727]: taffybar-linux-x86_64: xcb_io.c:259: poll_for_event: Assertion 
`!xcb_xlib_threads_sequence_lost' failed.                                                              
Apr 08 03:27:57 imalison-arch systemd[847]: taffybar.service: Main process exited, code=dumped, status=6
/ABRT                                                                                                  
Apr 08 03:27:57 imalison-arch systemd[847]: taffybar.service: Control process exited, code=exited status
=1                                                                                                     
Apr 08 03:27:57 imalison-arch systemd[847]: taffybar.service: Unit entered failed state.
Apr 08 03:27:57 imalison-arch systemd[847]: taffybar.service: Failed with result 'core-dump'.
Apr 08 03:27:58 imalison-arch systemd-coredump[2954]: Process 2727 (taffybar-linux-) of user 1000 dumped
 core.                                                                                                 
                                                      
                                                      Stack trace of thread 2727:
                                                      #0  0x00007f5f67f79a10 raise (libc.so.6)
                                                      #1  0x00007f5f67f7b13a abort (libc.so.6)
                                                      #2  0x00007f5f67f72607 __assert_fail_base (libc.so
.6)                                                                                                    
                                                      #3  0x00007f5f67f726b2 __assert_fail (libc.so.6)
                                                      #4  0x00007f5f6a2acc3a n/a (libX11.so.6)
                                                      #5  0x00007f5f6a2acceb n/a (libX11.so.6)
                                                      #6  0x00007f5f6a2ad135 _XReadEvents (libX11.so.6)
                                                      #7  0x00007f5f6a294d06 XIfEvent (libX11.so.6)
                                                      #8  0x00007f5f6c40050c gdk_x11_get_server_time (li
bgdk-x11-2.0.so.0)                                                                                     
                                                      #9  0x00000000004c4e4e egg_tray_manager_manage_scr
een (taffybar-linux-x86_64)                                                                            
                                                      #10 0x0000000000430834 taffybarzm0zi4zi6zmKMefjkBR
AJh55iNjqOkCix_SystemziTaffybarziSystray_systrayNew1_info (taffybar-linux-x86_64)                      
```

which is strange because the refreshTaffybar call has been wrapped with postGUIAsync

not sure why that is happening. any ideas?

travitch:
I'll have to look a bit more.  It seems very strange that the crash is in the tray manager.  Maybe it is referring to a thread started by the tray manager?

IvanMalison:
Well if I remove the tray it seems to avoid that crash, but its pretty easy to cause crashes for a whole bunch of reasons by just toggling over and over:

Here are a few more:

```
Apr 08 15:50:32 imalison-arch systemd-coredump[13547]: Process 12900 (taffybar-linux-) of user 1000 dump
ed core.                                                                                               
                                                       
                                                       Stack trace of thread 12900:
                                                       #0  0x00007f959b88fa10 raise (libc.so.6)
                                                       #1  0x00007f959b89113a abort (libc.so.6)
                                                       #2  0x00007f959b888607 __assert_fail_base (libc.s
o.6)                                                                                                   
                                                       #3  0x00007f959b8886b2 __assert_fail (libc.so.6)
                                                       #4  0x00007f959dbc2c3a n/a (libX11.so.6)
                                                       #5  0x00007f959dbc2ceb n/a (libX11.so.6)
                                                       #6  0x00007f959dbc2fdd _XEventsQueued (libX11.so.
6)                                                                                                     
                                                       #7  0x00007f959dbb4ca7 XPending (libX11.so.6)
                                                       #8  0x00007f959fd15575 n/a (libgdk-x11-2.0.so.0)
                                                       #9  0x00007f959e6251a9 g_main_context_check (libg
lib-2.0.so.0)                                                                                          
                                                       #10 0x00007f959e625744 n/a (libglib-2.0.so.0)
                                                       #11 0x00007f959e625b32 g_main_loop_run (libglib-2
.0.so.0)                                                                                               
                                                       #12 0x00007f95a00a03a7 gtk_main (libgtk-x11-2.0.s
o.0)                                                                                                   
                                                       #13 0x000000000042e734 taffybarzm0zi4zi6zmKMefjkB
RAJh55iNjqOkCix_SystemziTaffybar_defaultTaffybar13_info (taffybar-linux-x86_64)                        
```

```
Apr 08 15:50:32 imalison-arch env[12900]: [xcb] Unknown sequence number while processing queue
Apr 08 15:50:32 imalison-arch env[12900]: [xcb] Most likely this is a multi-threaded client and XInitThr
eads has not been called                                                                               
Apr 08 15:50:32 imalison-arch env[12900]: [xcb] Aborting, sorry about that.
Apr 08 15:50:32 imalison-arch env[12900]: taffybar-linux-x86_64: xcb_io.c:259: poll_for_event: Assertion
 `!xcb_xlib_threads_sequence_lost' failed.                                                             
```

```
Apr 08 15:49:12 imalison-arch env[11870]: The program 'taffybar-linux-x86_64' received an X Window Syste
m error.                                                                                               
Apr 08 15:49:12 imalison-arch env[11870]: This probably reflects a bug in the program.
Apr 08 15:49:12 imalison-arch env[11870]: The error was 'BadWindow (invalid Window parameter)'.
Apr 08 15:49:12 imalison-arch env[11870]:   (Details: serial 8771 error_code 3 request_code 40 minor_cod
e 0)                                                                                                   
Apr 08 15:49:12 imalison-arch env[11870]:   (Note to programmers: normally, X errors are reported asynch
ronously;                                                                                              
Apr 08 15:49:12 imalison-arch env[11870]:    that is, you will receive the error a while after causing i
t.                                                                                                     
Apr 08 15:49:12 imalison-arch env[11870]:    To debug your program, run it with the --sync command line
Apr 08 15:49:12 imalison-arch env[11870]:    option to change this behavior. You can then get a meaningf
ul                                                                                                     
Apr 08 15:49:12 imalison-arch env[11870]:    backtrace from your debugger if you break on the gdk_x_erro
r() function.)                                                                                         
```

I think that these are likely preexisting issues that are exascerbated by quick, repeated toggling.

simlu:
Possibly related to https://github.com/travitch/taffybar/issues/105?

IvanMalison:
> Possibly related to #105?

Nope, the errors from that almost always look like:

```
Apr 06 22:41:02 imalison-arch env[15506]: The program 'taffybar-linux-x86_64' received an X Window Syste
m error.                                                                                               
Apr 06 22:41:02 imalison-arch env[15506]: This probably reflects a bug in the program.
Apr 06 22:41:02 imalison-arch env[15506]: The error was 'BadWindow'.
Apr 06 22:41:02 imalison-arch env[15506]:   (Details: serial 20731 error_code 3 request_code 20 minor_co
de 0)                                                                                                  
Apr 06 22:41:02 imalison-arch env[15506]:   (Note to programmers: normally, X errors are reported asynch
ronously;                                                                                              
Apr 06 22:41:02 imalison-arch env[15506]:    that is, you will receive the error a while after causing i
t.                                                                                                     
Apr 06 22:41:02 imalison-arch env[15506]:    To debug your program, run it with the --sync command line
Apr 06 22:41:02 imalison-arch env[15506]:    option to change this behavior. You can then get a meaningf
ul                                                                                                     
Apr 06 22:41:02 imalison-arch env[15506]:    backtrace from your debugger if you break on the gdk_x_erro
r() function.)                                                                                         
Apr 06 22:41:02 imalison-arch systemd[811]: taffybar.service: Main process exited, code=exited, status=
```

simlu:
How is that significantly different?

IvanMalison:
> How is that significantly different?

Sorry I forgot that I posted that as 1 of the example bugs above. The point is that it is only 1 of a bunch of different crashes that can be triggered by rapidly toggling taffybar on a monitor.

IvanMalison:
@travitch false alarm on those bugs -- I think that I was accidentally running taffybar without my changes in #189 when I was initially testing the dbus stuff. I've been using it for over a week now, and things seem to work really well.

IvanMalison:
@travitch merge?

IvanMalison:
@travitch bump?

IvanMalison:
@travitch Is there something blocking merge here? I'm happy to make any changes that you'd like, but since multi-monitor support is such an often requested feature (this fixes #153, #183, #61, #139), I'd like to get this merged.

PR #187, opened 2017-04-06
Use bit operators for transforming pixel data in iconimages

IvanMalison:
This doesn't actually help all that much. The main issue is that we use a list representation at all, but fixing that is a bit tricky because the lists go pretty far down the call stack.


teleshoes:
@IvanMalison ACTUALLY, it appears this PR fixed `wshud` freezing+crashing for me.
i knew this was slow, but didnt realize how slow... thanks!

(yea, i impld it with bytelists because there were no libraries i could find and i wanted this test-less code to be as simple and clear as possible)

PR #189, opened 2017-04-09
Share A Single connection to the X11 Server/Display Object

### LINKED ISSUE #188: taffybar frequently crashes with `taffybar-linux-x86_64: user error (openDisplay)`
This crash occurs when we are attempting to open a new connection to the X11 server through xlib by calling [XOpenDisplay](https://tronche.com/gui/x/xlib/display/opening.html). The withDefaultCtx function (which is used all over the place seems to do this) which means we open a new connection for every function call that interacts with the X server.

This stack overflow question seems to indicate that it is better to call XOpenDisplay exactly once in your application:
http://stackoverflow.com/questions/14352098/best-practice-for-x11-application-one-xopendisplay-per-application-or-per-win

IvanMalison:
Fixes #188

### DIFF
--- a/src/System/Information/Battery.hs
+++ b/src/System/Information/Battery.hs
@@ -90,8 +90,8 @@ data BatteryInfo = BatteryInfo { batteryNativePath :: Text
 -}
                                }
 
--- | determin if a power source is a battery. The simple heuristic is
--- a substring search on 'BAT'.
+-- | determine if a power source is a battery. The simple heuristic is a
+-- substring search on 'BAT'.
 isBattery :: ObjectPath -> Bool
 isBattery = isInfixOf "BAT" . formatObjectPath
 
--- a/src/System/Information/X11DesktopInfo.hs
+++ b/src/System/Information/X11DesktopInfo.hs
@@ -19,10 +19,11 @@
 -----------------------------------------------------------------------------
 
 module System.Information.X11DesktopInfo
-  ( X11Context
+  ( X11Context(..)
   , X11Property
   , X11Window
   , withDefaultCtx
+  , getDefaultCtx
   , readAsInt
   , readAsListOfInt
   , readAsString
--- a/src/System/Taffybar/LayoutSwitcher.hs
+++ b/src/System/Taffybar/LayoutSwitcher.hs
@@ -26,7 +26,8 @@ module System.Taffybar.LayoutSwitcher (
   layoutSwitcherNew
 ) where
 
-import Control.Monad.IO.Class (MonadIO, liftIO)
+import Control.Monad.Trans
+import Control.Monad.Reader
 import qualified Graphics.UI.Gtk as Gtk
 import Graphics.X11.Xlib.Extras (Event)
 import System.Taffybar.Pager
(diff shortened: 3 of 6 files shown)


IvanMalison:
This extends #186

travitch:
This sounds like a good change - I'd just like to understand it a bit more before merging.

Is the implication that we were using multiple X connections (to the same display) before?  Do you suspect that this is related to the BadWindow errors we've been seeing for ages, or is this simply another bad X practice that may be causing different errors?

Is there some general principle we should be learning from this in how we interact with X?  Are there other places in the codebase we should be worrying about this?

IvanMalison:
> Is the implication that we were using multiple X connections (to the same display) before?

Yes. We initiate a new connection literally everywhere you can find a withDefaultCtx in the code right now.

> Do you suspect that this is related to the BadWindow errors we've been seeing for ages, or is this simply another bad X practice that may be causing different errors?

I have a good understanding of why the bad window errors occur:

A window can exist when some type of update is triggered, but not exist later when we request information about that window. Thats why we get a BadWindow error when two windows are closed (nearly) simultaneously. One window is closed, triggering an update, which makes us fetch the list of windows, which WILL include the second of the two windows that closes nearly simultaneously. By the time we request window about this second window it does not exist which leads to the BadWindow error.

> Is there some general principle we should be learning from this in how we interact with X?

Well, I think maybe what it indicates to me is that its time to take a look replacing the Xlib based x11 library that we are currently using with https://github.com/aslatter/xhb . Xlib is, IMO, a steaming pile of shit, and the way it handles errors is nearly impossible to deal with, especially when combined with the additional issues of dealing with the FFI with C. Its crazy that we can fully understand #105 and simply not be able to do anything about it because Xlib has such a terrible error handling system.


> Are there other places in the codebase we should be worrying about this?

I didn't get rid of uses of withDefaultCtx in WorkspaceSwitcher because it would have taken a pretty huge amount of work (Look at how much refactoring I had to do in WorkspaceHUD), and I really think that WorkspaceHUD should eventually replace it (since it has complete feature parity with WorkspaceSwitcher). @teleshoes claimed that he still experiences bugs with WorkspaceHUD, and its hard for me to claim that they don't exist when I get crashes frequently (whether I use HUD or Switcher).

I think that almost all of the bugs are X11 related, but they are mostly very subtle and its not clear how to fix them. Again, I think that the answer might be to look at https://github.com/aslatter/xhb

IvanMalison:
> This sounds like a good change - I'd just like to understand it a bit more before merging.

I've been using it for a while now, but it would definitely benefit from a bit of testing before merge. It can sometimes be hard for me to tell if the crashes I'm getting are new or not, because some of the crashes cause spurious, unrelated core dumps because they happen asynchronously.

@sakshamsharma @teleshoes @escherdragon any help testing would be appreciated.

IvanMalison:
@travitch ping?

djvs:
This isn't available on cabal yet, I take it?

IvanMalison:
@djvs I imagine we will do another version bump for whatever makes it into this request. Probably 0.5.0 given the significance of changes on master.

Not sure what @travitch thinks.

IvanMalison:
@djvs its not really that hard to install taffybar from source though. Thats what I would recommend anyway.

djvs:
Wow, loving the new version so far.

djvs:
Still getting this error from git taffybar though:

taffybar-linux-x86_64: user error (openDisplay)

IvanMalison:
@djvs Really? Did you use --force-reinstalls when you ran cabal install? Are you sure you are using the newest version.

djvs:
I'm using the pager, if that's the same as the WorkspaceSwitcher.  But no, I didn't do --force-reinstalls, so I'll report back if I see another crash after doing so.

IvanMalison:
@djvs no using the pager is the problem 100%, can you just trying using

```
taffyPagerHUDLegacy :: PagerConfig -> IO Widget
taffyPagerHUDLegacy cfg = do
  pgr <- pagerNew cfg
  whud <- buildWorkspaceHUD (hudFromPagerConfig cfg) pgr
  los <- layoutSwitcherNew pgr
  wnd <- windowSwitcherNew pgr
  sp1 <- separator cfg
  sp2 <- separator cfg
  box <- hBoxNew False 0

  boxPackStart box whud PackNatural 0
  boxPackStart box sp1 PackNatural 0
  boxPackStart box los PackNatural 0
  boxPackStart box sp2 PackNatural 0
  boxPackStart box wnd PackNatural 0

  widgetShowAll box
  return (toWidget box)
```

instead of `taffyPagerNew`

IvanMalison:
you'll need to import
``` haskell
import System.Taffybar.WorkspaceHUD
```

djvs:
After patching up all those unqualified imports:

```

taffyPagerHUDLegacy :: PagerConfig -> IO Gtk.Widget
taffyPagerHUDLegacy cfg = do
  pgr <- pagerNew cfg
  whud <- buildWorkspaceHUD (hudFromPagerConfig cfg) pgr
  los <- layoutSwitcherNew pgr
  wnd <- windowSwitcherNew pgr
  sp1 <- separator cfg
  sp2 <- separator cfg
  box <- Gtk.hBoxNew False 0

  Gtk.boxPackStart box whud Gtk.PackNatural 0
  Gtk.boxPackStart box sp1 Gtk.PackNatural 0
  Gtk.boxPackStart box los Gtk.PackNatural 0
  Gtk.boxPackStart box sp2 Gtk.PackNatural 0
  Gtk.boxPackStart box wnd Gtk.PackNatural 0

  Gtk.widgetShowAll box
  return (Gtk.toWidget box)
```

Left wondering what "separator" is?

IvanMalison:
> Left wondering what "separator" is?

It's a field in the config. I think its string/widget that is placed between the main widgets. Not really important.

djvs:
Well, it compiles without those 4 lines.  Tried to copy the separator definition from here:

https://hackage.haskell.org/package/taffybar-0.3.0/src/src/System/Taffybar/TaffyPager.hs

which did work except for some type annotation issue.  Going without them for now.  Thanks!

IvanMalison:
> Well, it compiles without those 4 lines. Tried to copy the separator definition from here:
> 
> https://hackage.haskell.org/package/taffybar-0.3.0/src/src/System/Taffybar/TaffyPager.hs
> 
> which did work except for some type annotation issue. Going without them for now. Thanks!
>  Delete branch

@djvs Eventually you should just use `taffyPagerHUDNew`,

but that will require you to come up with the HUDConfig for your workspaceHUD (see 
https://github.com/travitch/taffybar/blob/788463f266ec74b277b9683348b2a6f5138c4224/src/System/Taffybar/WorkspaceHUD.hs#L127)

I would recommend you try adding Icons with workspaceHUD, they're pretty awsome.

djvs:
https://gist.github.com/anonymous/a70f261bcf02dc82ee043e1122adc88a

My current taffybar.hs - there are icons next to my workspaces already: 



Pretty nice.

PR #200, opened 2017-05-15
Include TaffyPager example in main docs

frioux:
This seems like the recommendation going forward, so I would consider putting it
*before* the stdin version, but I wasn't sure.

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -70,6 +70,47 @@ module System.Taffybar (
   -- They are actually [IO Widget] since the bar needs to construct them
   -- after performing some GTK initialization.
 
+  -- * XMonad Integration (via EWMH hints and X11 events)
+  --
+  -- | The TaffyPager widget differs from its xounterpart in xmobar: it listens
+  -- for updates using EWMH hints and X11 events instead of reading from stdin.
+  -- See [System.Taffybar.TaffyPager] for more details.  First, you'll need to
+  -- configure XMonad to advetize the EWMH hints and X11 events:
+  --
+  -- > import XMonad.Hooks.EwmhDesktops (ewmh)
+  -- > import XMonad.Hooks.ManageDocks
+  -- > import System.Taffybar.Hooks.PagerHints (pagerHints)
+  -- >
+  -- > main = do
+  -- >   xmonad $ ewmh $ pagerHints $ defaultConfig { manageHook = manageDocks
+  -- >                                              }
+  --
+  -- And then your taffybar configuration can simply be:
+  --
+  -- > import System.Taffybar
+  -- > import System.Taffybar.Systray
+  -- > import System.Taffybar.TaffyPager
+  -- > import System.Taffybar.SimpleClock
+  -- > import System.Taffybar.Widgets.PollingGraph
+  -- > import System.Information.CPU
+  -- >
+  -- > cpuCallback = do
+  -- >   (_, systemLoad, totalLoad) <- cpuLoad
+  -- >   return [ totalLoad, systemLoad ]
+  -- >
+  -- > main = do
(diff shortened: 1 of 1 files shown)


frioux:
I don't think my changes caused that failure; if they did please let me know how.

src/System/Taffybar.hs
```diff
@@ -70,6 +70,49 @@ module System.Taffybar (
...
+  -- for updates using EWMH hints and X11 events instead of reading from stdin.
+  -- See [System.Taffybar.TaffyPager] for more details.  First, you'll need to
+  -- configure XMonad to advetize the EWMH hints and X11 events:
+  --
+  -- > import XMonad.Hooks.EwmhDesktops (ewmh)
+  -- > import XMonad.Hooks.ManageDocks
+  -- > import System.Taffybar.Hooks.PagerHints (pagerHints)
+  -- >
+  -- > main = do
+  -- >   client <- connectSession
```
IvanMalison: What does this do? I don't see any uses of client in what follows.

src/System/Taffybar.hs
```diff
@@ -70,6 +70,49 @@ module System.Taffybar (
...
+  -- >   let pp = defaultPP
```
IvanMalison: I'm pretty sure this is not used anywhere

frioux:
@IvanMalison I amended my original commit; I could have swore I removed that but clearly I forgot.  Good catch and thanks

src/System/Taffybar.hs
```diff
@@ -70,6 +70,47 @@ module System.Taffybar (
   -- They are actually [IO Widget] since the bar needs to construct them
   -- after performing some GTK initialization.
 
+  -- * XMonad Integration (via EWMH hints and X11 events)
+  --
+  -- | The TaffyPager widget differs from its xounterpart in xmobar: it listens
```
IvanMalison: xounterpart -> counterpart?

src/System/Taffybar.hs
```diff
@@ -70,6 +70,47 @@ module System.Taffybar (
...
+  -- for updates using EWMH hints and X11 events instead of reading from stdin.
+  -- See [System.Taffybar.TaffyPager] for more details.  First, you'll need to
+  -- configure XMonad to advetize the EWMH hints and X11 events:
+  --
+  -- > import XMonad.Hooks.EwmhDesktops (ewmh)
+  -- > import XMonad.Hooks.ManageDocks
+  -- > import System.Taffybar.Hooks.PagerHints (pagerHints)
+  -- >
+  -- > main = do
+  -- >   xmonad $ ewmh $ pagerHints $ defaultConfig { manageHook = manageDocks
```
IvanMalison: This is outdated, the docks wrapper for xmonad should be used instead.

src/System/Taffybar.hs
```diff
@@ -70,6 +70,47 @@ module System.Taffybar (
...
+  -- >   (_, systemLoad, totalLoad) <- cpuLoad
+  -- >   return [ totalLoad, systemLoad ]
+  -- >
+  -- > main = do
+  -- >   let cpuCfg = defaultGraphConfig { graphDataColors = [ (0, 1, 0, 1), (1, 0, 1, 0.5)]
+  -- >                                   , graphLabel = Just "cpu"
+  -- >                                   }
+  -- >       clock = textClockNew Nothing "<span fgcolor='orange'>%a %b %_d %H:%M</span>" 1
+  -- >       pager = taffyPagerNew defaultPagerConfig
+  -- >       tray = systrayNew
```
IvanMalison: lets make this configuration as minimal as possible to avoid any confusion about what does an does not need to be there. I think just having the pager and the tray (since they require no config) would be good.

frioux:
I made this PR 10 months ago and all the context has left my brain.  I made it because I found it hard to extend TaffyBar more than in the most basic ways.  Feel free to close this, or take it and make all the changes you want.

PR #65, opened 2014-01-31
Added basic volume control

abdulsattar:
``` haskell
import System.Taffybar.Volume
...
volume = volumeTextNew "default" "Master" 10
```

Shows the ceiled percentage of volume.

PR #203, opened 2017-06-03
Volume control (continued from #65)

NickHu:
Conflicts resolved and brought up to date with master.


travitch:
Thanks.  I don't know if there are any non-Linux users, but do you think it would be possible to only conditionally enable this module on Linux (given the ALSA dependency)?

NickHu:
I have no idea I'm afraid, and I don't have any systems running other operating systems to test.
A quick googling seems to suggest that cabal can condition on operating systems - if you added that into the cabal file I wouldn't object.

IvanMalison:
@NickHu Can you handle the merge conflicts?

NickHu:
Done

PR #204, opened 2017-06-03
Add Icon widgets

NickHu:
I wrote this a while ago, and have been using it to display static and dynamic icons in taffybar.

For an example of a dynamic icon, consider an icon which indicates whether a laptop is on AC or battery:

in `taffybar.hs`

```haskell
batteryIconFunc :: BatteryContext -> String -> IO String
batteryIconFunc ctxt prefix = do
  ac <- batteryAC ctxt
  percentage <- batteryPercent ctxt
  case () of
    _ | ac == BatteryStateCharging -> return $ prefix ++ "ac_01.xpm"
      | percentage < 0.1 -> return $ prefix ++ "bat_empty_02.xpm"
      | percentage < 0.4 -> return $ prefix ++ "bat_low_02.xpm"
      | otherwise -> return $ prefix ++ "bat_full_02.xpm"

batteryPercent :: BatteryContext -> IO Double
batteryPercent ctxt = do
  info <- getBatteryInfo ctxt
  return (batteryPercentage (fromJust info) / 100)

batteryAC :: BatteryContext -> IO BatteryState
batteryAC ctxt = do
  info <- getBatteryInfo ctxt
  return . batteryState $ fromJust info
```

where `prefix` is something like `/home/nick/icons/`

I don't know if this duplicates functionality of System.Taffybar.IconImages, because I can't understand what that does - feedback appreciated.

### DIFF
--- a/src/System/Taffybar/Widgets/Icon.hs
+++ b/src/System/Taffybar/Widgets/Icon.hs
@@ -1,0 +1,58 @@
+-- | This is a simple static image widget, and a polling image widget that
+-- updates its contents by calling a callback at a set interval.
+module System.Taffybar.Widgets.Icon
+(
+  iconImageWidgetNew,
+  pollingIconImageWidgetNew
+) where
+
+import Control.Concurrent ( forkIO, threadDelay )
+import Control.Exception as E
+import Control.Monad ( forever )
+import Graphics.UI.Gtk
+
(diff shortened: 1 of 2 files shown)


travitch:
It looks like `System.Taffybar.IconImages` just loads image files into pixbufs, and isn't providing a widget.  Thanks for the contribution.

PR #208, opened 2017-07-21
add optional activeLayoutIO to pagerConfig (non-breaking change)

teleshoes:
non-breaking change to active-layout to allow further customizing layout with markup.

i use this to add a count of windows in full-screen-mode "[5]"
```
activeLayoutIO   = \layout -> case layout of
      "left"    -> return "[]="
      "top"     -> return "TTT"
      "grid"    -> return "###"
      "full"    -> do
        cnt <- windowCount
        let numFmt = if 0 <= cnt && cnt < 10 then show cnt else "+"
        let color = if cnt > 1 then fgbg "blue" "red" else id
        return $ color $ "[" ++ numFmt ++ "]"
```

### DIFF
--- a/src/System/Taffybar/LayoutSwitcher.hs
+++ b/src/System/Taffybar/LayoutSwitcher.hs
@@ -77,8 +77,8 @@ layoutSwitcherNew pager = do
 pagerCallback :: PagerConfig -> Gtk.Label -> Event -> PagerIO ()
 pagerCallback cfg label _ = do
   layout <- liftPagerX11 $ readAsString Nothing xLayoutProp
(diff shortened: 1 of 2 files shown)


teleshoes:
@travitch closed and re-opened because it was ancient and had merge conflicts

teleshoes:
@travitch merge conflicts fixed, again ;)

IvanMalison:
I'm going to merge this for now, but we already have breaking changes in pagerConfig, so I think we should consider getting rid of activeLayout

teleshoes:
woo, thanks!
(by all means, its silly to have both if we dont have to worry about breakage)

PR #209, opened 2017-07-21
workspace-images: fix preferCustomIcon

teleshoes:
tested in `WSS` and `WSHUD`

the old prefer-custom-icon behaviour makes `preferCustomIcon=True` useless, as it always overrides the EWMH icon if you use an "unknown" icon.

this replaces the boolean `preferCustomIcon` with a boolean `hasIcon` param to `customIcon`, letting the pager config handle the situation intelligently.

this also allows case-by-case basis for preferring custom icon.
(i use this to override "unknown" icon and also certain websites in firefox, and otherwise prefer system icon instead of custom)

for folks that use custom icon, this requires tiny change to pagerConfig. if you use system icons only, no change required.
`customIcon t c -> fct t c,` => `customIcon _ t c -> fct t c,`

### DIFF
--- a/src/System/Taffybar/Pager.hs
+++ b/src/System/Taffybar/Pager.hs
@@ -92,10 +92,8 @@ data PagerConfig = PagerConfig
   , imageSize               :: Int
   -- ^ fill empty images instead of clearing them
   , fillEmptyImages         :: Bool
-  -- ^ use custom icons over EWHMIcons
-  , preferCustomIcon        :: Bool
-  -- ^ get icon based on window title and class
-  , customIcon              :: String -> String -> Maybe FilePath
+  -- ^ get custom icon based on: has-EWMH-icon, window-title, window-class
+  , customIcon              :: Bool -> String -> String -> Maybe FilePath
   -- ^ title windows for WindowSwitcher
   , windowSwitcherFormatter :: M.Map WorkspaceIdx String -> X11WindowHandle -> String
   }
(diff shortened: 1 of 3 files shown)


IvanMalison:
kinda a bummer that we have to change the interface here. I also think that Ideally the customIcon function would actually take in as an argument whatever the current icon is instead of taking in a bool. That way you get rid of the conditional logic around whether there is a customIcon/ewmhIcon, and simply always return the result of the customIcon function (which can just return whatever it was passed in if it does not want to override).

teleshoes:
yea, i tried to work around breaking api, or add a parameter and have the old ones call the new one with some zany function, but it was messy as hell, and i doubt many people use the customIcon functor. the break is at least very easy to fix.

i considered passing in the icon, but whats the point of passing around a bytestring or a pixbuf or what-have-you? the user isnt going to unwrap it anyway, beyond calling isJust.
assuming thats true, then this is equivalent (the user returns None to not override).

(also, we cant use the output of the custom-image-selector because it just returns a filepath. leaving the pixbuf unwrapping to the enduser is worse, IMO)

IvanMalison:
> i considered passing in the icon, but whats the point of passing around a bytestring or a pixbuf or what-have-you? the user isnt going to unwrap it anyway, beyond calling isJust.
> assuming thats true, then this is equivalent (the user returns None to not override).

The code is just tidier/idiomatic that way IMO. Also in the case of WSHUD it allows for easier composition. You could wrap one custom function with another without having to worry about the details. Maybe one of these functions would only override EWMH icons, but not custom filepaths. You wouldn't need an extra field to indicate that there is another overrding custom file etc.

IvanMalison:
@teleshoes Resolve merge conflicts?

teleshoes:
@IvanMalison resolved. conflict was with windowSwitcherFormatter, and a function that moved higher up

IvanMalison:
@travitch Merge? This looks pretty good to me.

PR #211, opened 2017-07-27
Improved the FreedesktopNotifications widget

drdo:
Instead of having a separate current notification in noteCurrent, we can just take the current notification to be the first one in noteQueue. This greatly simplifies the code.

Allowed notifications to never timeout by wrapping notificationMaxTimeout in a Maybe.
Set Nothing to be the default.

We can pass the whole sequence to the formatter to do interesting things with it.
Changed the default formatter so that it displays the number of pending notifications.

The displayThread now only handles displaying the notification, in particular the timeout thread is started by notify instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
  * Quick start script for installing from git with stack (Ivan Malison)
  * Add a volume widget (Nick Hu and Abdul Sattar)
  * Add available memory field to MemoryInfo (Will Price)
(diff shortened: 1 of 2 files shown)


IvanMalison:
@drdo @u11gh can you guys fix the merge conflicts?

u11gh:
Looks like the conflicts can be solved easily.

drdo:
Conflicts fixed.

PR #212, opened 2017-07-28
Add a customization parameter to pager for window switcher titles

IvanMalison:


travitch:
I'm not familiar with the `X11WindowHandle` type - would it be useful to have the `windowSwitcherFormatter` hook be in `IO` instead of pure?  Are there useful pure operations on `X11WindowHandle`?

travitch:
Just for context, I'm in favor of the change - I just want to avoid breaking the API later when/if we decide that purity hurts there.

IvanMalison:
> 'm not familiar with the X11WindowHandle type - would it be useful to have the windowSwitcherFormatter hook be in IO instead of pure?

Possibly. X11Property might actually be better than IO. I thought about this, but decided I wanted to avoid making too many changes.

> Are there useful pure operations on X11WindowHandle?

Basically just selecting which of the workspace, window class and window title you want to display and in what order.



> Just for context, I'm in favor of the change - I just want to avoid breaking the API later when/if we decide that purity hurts there.

Yeah, generally speaking, I think that purity for these customization features can end up being a pain because its hard to foresee every single last piece of information that might be useful in the customization context.

PR #214, opened 2017-08-28
Add available memory field to MemoryInfo

willprice:
On my system, `cat /proc/meminfo` yields:

```
MemTotal:       16289508 kB
MemFree:         1059928 kB
MemAvailable:    6774876 kB
Buffers:          954708 kB
Cached:          2824500 kB
SwapCached:           28 kB
...
```

Using the previously calculation of free memory `rest = free + buffer + cache`
yields 4839136 kB, which is substantially different to `MemAvailable`. I
wish to chart `(MemTotal - MemAvailable)/MemTotal` as a more accurate
statistic of how much RAM on my system is being used

### DIFF
--- a/src/System/Information/Memory.hs
+++ b/src/System/Information/Memory.hs
@@ -14,25 +14,27 @@ data MemoryInfo = MemoryInfo { memoryTotal :: Double
                              , memorySwapFree :: Double
                              , memorySwapUsed :: Double -- swapTotal - swapFree
                              , memorySwapUsedRatio :: Double -- swapUsed / swapTotal
+                             , memoryAvailable :: Double -- An estimate of how much memory is available for starting new apps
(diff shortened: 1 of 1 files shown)


src/System/Information/Memory.hs
```diff
@@ -10,23 +10,25 @@ data MemoryInfo = MemoryInfo { memoryTotal :: Double
                              , memoryFree :: Double
                              , memoryBuffer :: Double
                              , memoryCache :: Double
+                             , memoryAvailable :: Double -- An estimate of how much memory is available for starting new applicaitons
```
teleshoes: typo, applicaitons

IvanMalison: @willprice can you fix this and the merge conflicts?

PR #216, opened 2017-08-29
Call initThreads in main

### LINKED ISSUE #201: Unknown sequence number while processing queue
```
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
taffybar-linux-x86_64: xcb_io.c:259: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
```

IvanMalison:
Fixes #201

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -165,12 +165,12 @@ module System.Taffybar (
   -- >       _ <- Gtk.on ebox Gtk.buttonPressEvent systemEvents
   -- >       Gtk.widgetShowAll ebox
   -- >       return $ Gtk.toWidget ebox
-  -- 
+  --
   -- The resulting widget can be used like normal widgets, but you can use
   -- different mouse buttons to run various programs and it has a useful tooltip
   -- which shows the concrete numbers, which may not be clear in the graph
   -- itself.
-  
+
   TaffybarConfig(..),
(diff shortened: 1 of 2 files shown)


IvanMalison:
This may improve stability in a bunch of other places as well, but more investigation is needed to confirm whether we can close additional issues.

travitch:
Do we know if the Haskell X11 library (or GTK) has appropriate calls to `XLockDisplay` and `XUnlockDisplay`?  This seems like a possibly reasonable thing for now, though the whole story isn't clear to me yet.  Longer term, we probably need to re-architect things to avoid all multi-threaded calls to GTK unless we can figure out a guaranteed safe approach.

IvanMalison:
> Do we know if the Haskell X11 library (or GTK)

GTK certainly does NOT which is why it has a 'UI' thread, and we need to do everything with postGUIAsync.

We'll need https://github.com/xmonad/X11/pull/55 to get merged in and released before we are even able to call XUnlockDisplay in any of our code, but we can start doing that then.

For now, Simply calling XInitThreads does fix a couple of problems. I am a little bit scared that its not clear WHY it does that, but I suspect that Xlib may internally guard against certain things when XInitThreads is called

PR #218, opened 2017-08-29
Full stack support, quick start script for quick install

IvanMalison:
Also includes some misc style fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 language: c
 cache:
   directories:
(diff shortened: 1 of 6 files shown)


IvanMalison:
I guess this change is not compatible with ghc 7.8 because dyre is not compatible with that version. Is there any reason we shouldn't drop support?

IvanMalison:
Though the pull request build fails, it will pass once the relevant sha is part of this repo (see https://travis-ci.org/IvanMalison/taffybar/builds/270124937)

PR #221, opened 2017-09-07
My PRs combined

IvanMalison:
This combines #220 #218 #212 #210 since they have non trivial merges

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 language: c
 cache:
   directories:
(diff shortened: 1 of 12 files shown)


IvanMalison:
yes afaict, things suddenly stopped working with addons/apt/package. If i remove the sudo: required and the apt-get the build fails saying that sutff is missing. I tried messing around with it a bit and I just couldn't get it to work.



> I'm also not hugely fond of the git/dyre stuff, but I won't hold this up over it...

Well I actually didn't end up changing anything in the main stack.yaml files. This simply adds a new stack.yaml.example (template/example file) that shows how to get a stack build ones taffybar with stack.

travitch:
Fair enough, thanks.

PR #222, opened 2017-09-11
Allow effects in label setter

IvanMalison:


travitch:
When I can find some time, I think I'm going to go through and allow side effects in most of the formatters that are already defined.  It seems to come up regularly.

Issue #228, opened 2017-10-03
Taffybar GTK Window Stops Refreshing (hangs) in WSHUD


teleshoes:
the ui sometimes becomes unresponsive when i use wshud instead of wss. it happens seemingly at random, and usually after a full day of uptime, but it happens regularly enough that im certain it only happens in wshud.

its likely a deadlock in the ui or some other concurrency related problem preventing re-painting. clicking things still works, its just that labels and images never change (all widgets, the wshud, the clock, cpu graphs, etc)

i would blame my taffybar, but if i change just one line to switch to WSS, the problem goes away. (i was using WSS for months and then i switched to WSHUD a coupla weeks ago and it started dying regularly again).

i havent (re-?)reported ths bug because i havent been on the LATEST wshud code, but i am now, and it happened twice in the last 3 days. im still trying to collect more data, but i wanted to put this out here now jic

teleshoes:
@IvanMalison 
one more thing; i used to think that this was caused by the slow-image-loading problem, but thats definitely fixed by the latest code.

(changing images quickly doesnt freeze the ui at all, and the image updates as fast as the windows switch. before the latest image-loading commit, the images loaded fairly fast, but they still blocked the ui updating for a tiny bit. further back, before your fix last year, it was really slow AND blocked the ui)

IvanMalison:
@teleshoes Can you give details about which version of ghc you are using? What about your X11 version? What distro are you running? We're getting to the point here with this that I think we have to start looking to these things as potential causes of your issues. Can you build with stack so all of this is 100% reproducible? Does RSS usage go up? Can you post your config so I can run it for a couple of days to see if I can reproduce?

BTW: I have a really hard time believing that you are able to run WSS for multiple days at a time since 
#188 #105 are not fixed in WSS. When I run WSS on my machine I am lucky to get more than an hour of uptime without one of those two causing a crash.

> that labels and images never change

So you are saying it is 100% unresponsive? Or it is it just slow? I can't quite tell from this report.

> slow-image-loading problem,

What is this referring to? I'm not aware of having fixed that. Images are still really slow because they are loaded into lists and not arrays.

> still blocked the ui

Based on this it seems like you are reffing to the rate limiting changes that I added to prevent image updates from hogging cpu?

teleshoes:
> Can you give details about which version..

ghc 8.0.2
xorg 7.7+19
xserver 1.19.3-2
debian sid
linux 4.12.8
xmonad 0.13

[...]

PR #229, opened 2017-10-14
Safer X11, Faster Icon Performance

### LINKED ISSUE #105: Crash when multiple windows close simultaneously

### LINKED ISSUE #191: Icon Performance is really bad with certain icons
Spotify and Atom are both examples of icons that have very bad performance. Their performance is so bad because they don't provide low resolution icons, and because we currently use (linked) lists to represent them in memory. They need to remain contiguous in memory from the point where they are fetched from x11

IvanMalison:
I finally figured #105 out! Unfortunately it took custom bindings to a few xlib calls, and our own dedicated x11 event loop to do it...

We'll want to monitor xmonad/X11#51 to see if what they end up coming up with there is suitable for our needs. Theres a lot of work to be done there though, so I don't think waiting for their solution is a viable option, especially since this fixes the single most annoying outstanding bug in taffybar.

Because I ended up writing our own bindings to getProp I also made it so that the icon code doesn't use haskell lists anymore, which DRASTICALLY improves icon performance. This pretty much 100% resolves #191 for me. For anyone wondering how bad things WERE try running spotify with icons on before this change and you'll see that there is massive lag whenever that icon needs to be displayed or updated.

I don't have my hopes too high, but this could also have an impact on #228, since so many things were changed. 

IMPORTANT: I actually can't guarantee that this change fixes #105 for workspace switcher, you'll have to switch to WorkspaceHUD if you want to be assured that this is fix. That said, it should just be a matter of running all x11 calls through the dedicated thread. I know that @teleshoes has strong feelings about this, but I really think that we should consider deprecating WSS at this point.


simlu:
Omg, I'm so excited for this! I can finally get rid of my auto-restart script!

IvanMalison:
> Omg, I'm so excited for this! I can finally get rid of my auto-restart script!

Awesome

@simlu Do you have a consistent repro/are you able to install from source to verify that this does, indeed fix the issue for you?

I have to admit that while I mostly think I understand what is going on, there are one or two details that are a little hazy for me right now, and I want to make sure that this works for others before declaring the issue resolved.

simlu:
I won't be able to do any testing until later this month unfortunately. Have you tried running my script that I posed in #105?

I use VMs and can test from source, but usually revert afterwards since I dislike having those various dependencies installed.

IvanMalison:
@simlu Oh no! It looks like it is SOMETIMES able to handle your script and sometimes not. Pretty weird, right? If I reduce the number of windows involved it's a lot more consistent. I think that maybe I missed a spot that normally runs very quickly, or perhaps it has to do with window switcher.

The good news is that it does log that it encounters an error before dying. I'm so bummed though. I'm going to look in to it a bit more though.

travitch:
Thanks for looking at this.  My read of this is that there were roughly two issues:
 * Some functions from the X11 wrapper can throw exceptions that weren't being caught and were killing threads
 * While we seem to mostly do an okay job with single-threaded access to GTK, we didn't do so well with raw X11 functions and were doing thread-unsafe things with X11

Do we basically need to audit everything we have in the X11 monad?  Should we look at a solution that prevents this more thoroughly and perhaps remove all unsafe functions?

simlu:
@IvanMalison Is there any change in the amount of crashes using my config? I've rewritten a bunch of code that did unsafe access. Did you try stripping the bar config to an absolute minimum?

Also what is the error that is being logged?

@travitch This is what I did with my config basically (I think). I'm curious if that plus the changes @IvanMalison made potentially fixes it.

IvanMalison:
> Some functions from the X11 wrapper can throw exceptions that weren't being caught and were killing threads

Not exactly. A few things conspired to cause #105

- `xGetWindowProperty` and other X11 funtions were marked as unsafe which meant that anything interrupting BACK to haskell would cause the entire runtime to be frozen. [This post](https://www.reddit.com/r/haskell/comments/1vsw1z/how_to_work_around_blocking_in_an_unsafe_ffi_call/cew18mv/) explains it well if you are curious.
- We weren't properly setting an X11Error handler (which is very different than catching errors in haskell), so when an error occurred it would crash the program.
- Previously, when I tried properly setting an X11Error handler, it would fail because the provided X11 handler was a function in the haskell runtime (which is why the first issue mentioned is important)
- Less significant, but still a hurdle that I had to overcome in this pull request: A lot of X11 calls were actually being made on the UI thread for Gtk in a blocking way which was bad because that can cause the UI thread to hang

@travitch @simlu
The key point here is that most of these exceptions CANT be caught in any way in haskell, especially with how the FFI was configed. The solution to the main problem was to BOTH set a handler AND mark the FFI calls as safe. This caused an issue with X11 requests and threads, which is why I made a dedicated thread to handle all X11 requests.

> While we seem to mostly do an okay job with single-threaded access to GTK, we didn't do so well with raw X11 functions and were doing thread-unsafe things with X11

Yes, although, in theory, X11 should be usuable from different threads when XInitThreads is called. For whatever reason, when I marked the X11 functions as safe, I started getting a bunch of errors about unknown request, even though we DO initialize the threading system of X11, which should (in theory) allow for calls to X11 from different threads.

> Do we basically need to audit everything we have in the X11 monad?

Yes, possibly, although currently, at least with the things touched by my own configuration, I think I've been able to stamp out all of these. Also, it is possible to call in to X11 without using that monad, and we do in some places, IIRC.

> Should we look at a solution that prevents this more thoroughly and perhaps remove all unsafe functions?

That's basically what I've done, although I just realized that I missed getGeometry, which is very likely the cause of the remaining issues that I am seeing.

IvanMalison:
Great news! It turns out that the reason there were still some crashes WAS that `getGeometry` was not safe!

@simlu 
> @IvanMalison Is there any change in the amount of crashes using my config?

Yes. I am actually now able to run your script and avoid a crash every time.

> I've rewritten a bunch of code that did unsafe access. Did you try stripping the bar config to an absolute minimum?

It seems like you mostly just added `catchAny` in a whole bunch of places. As I mentioned [above](https://github.com/travitch/taffybar/pull/229#issuecomment-336615532), the exceptions from Xlib can't be caught in Haskell, so I don't think that any of the rewrites that you've done are doing anything, unfortunately.

> I'm curious if that plus the changes @IvanMalison made potentially fixes it.

I think that the issue is 100% fixed assuming that the only widget you are using is WorkspaceHUD. I'm going to look at the others in a little bit, but I'd love it if you could validate that running just WorkspaceHUD can handle the issue.

IvanMalison:
@simlu @travitch Here's a gif of the new code handling @simlu's script from #105

teleshoes:
@IvanMalison woo, nice stuff man!

if this DOES fix 228, i would absolutely agree with deprecating WSS. WSHUD is easier to use and extend, has more features, and is more flexible; if only WSHUD didnt crash all the time for me.

i used WSHUD for the past week, and it crashed at least twice a day every workday.

i have switched to 229 (plus cherry-picked 208+209 on top). ill let you know after a few workdays go by how 228 stands.

IvanMalison:
@teleshoes If you do get crashes, can you check whether they are due to #231, and then see if you ALSO get the issues in WSS when applying this PR?

I'm a little bit worried that I may have caused a double free somewhere with the refactors to the icon image code.

teleshoes:
will do, though im pretty sure ive never seen that exact crash before

travitch:
It sounds like this is a significant improvement over what we had before, so I'm going to go ahead and merge it.  We can fix any issues that pop up over the next week, hopefully.  If it looks stable for a week, I'll go ahead and cut a release.

IvanMalison:
@travitch Awesome. Things have been pretty stable for me so far, but I do want to monitor #231 just to make sure that it isn't something that was introduced by this change.

Great news about cutting a release! I think We're at a good point for doing something like that!

PR #232, opened 2017-10-16
Add 'usePrimaryMonitor' function for 'getMonitorConfig' option

iamjamestl:
This function uses Xrandr library to determine the primary monitor number.  If a primary monitor isn't set or otherwise can't be determined, the function falls back to the `monitorNumber` config option.

See: https://github.com/travitch/taffybar/pull/155

I debated about whether to add this functionality to existing modules or implement it separately, like `System.Taffybar.ToggleMonitor`.  I decided to implement it in the existing modules because it seems to fit in and it seems like a basic option that users would expect to be built-in.

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -179,6 +179,7 @@ module System.Taffybar (
   taffybarMain,
   allMonitors,
   useMonitorNumber,
+  usePrimaryMonitor,
   ) where
 
 import qualified Config.Dyre as Dyre
@@ -194,6 +195,7 @@ import Safe ( atMay )
 import System.Environment.XDG.BaseDir ( getUserConfigFile )
 import System.Exit ( exitFailure )
 import System.FilePath ( (</>) )
+import System.Information.X11DesktopInfo
 import qualified System.IO as IO
 import System.Mem.StableName
 import Text.Printf ( printf )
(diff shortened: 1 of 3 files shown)


src/System/Taffybar.hs
```diff
@@ -258,6 +260,13 @@ useMonitorNumber c@(cfg, _) = return umn
             | mnumber == monitorNumber cfg = Just c
             | otherwise = Nothing
 
+-- | Use the primary monitor as set by Xrandr.
+usePrimaryMonitor :: TaffybarConfigEQ -> IO (Int -> Maybe TaffybarConfigEQ)
+usePrimaryMonitor c@(cfg, _) = do
+  maybePrimary <- withDefaultCtx getPrimaryOutputNumber
```
IvanMalison: Just a note: using withDefaultCtx too liberally, which actually opens a new X11 connection is what led to #188, and forced me to write #189.

In this case, I think its fine, since this function will only be called at startup, and any time taffybar recalculates which monitors it should display on, which is not going to be very often. You don't really have a way to access the pager x11 connection anyway, so not much can be done, I suppose.

IvanMalison:
Mark this as closing #155. LGTM

IvanMalison:
@travitch Merge?

PR #233, opened 2017-10-16
WorkspaceHUD: Fix urgent state + export WorkspaceState

iamjamestl:
This is a pair of small commits that I can break out into separate PRs if desired.

1. Export `WorkspaceState` data type so that custom `labelSetter` functions can be defined based on the workspace state (active, urgent, etc.).

2. Set the urgent state only on workspaces containing urgent windows.  This fixes (imo) a bug where all of the workspaces would be marked urgent, even if only one of them contained an urgent window.  I've also prioritized the active and visible state over the urgent state, which is how the WorkspaceSwitcher widget behaves, namely that I only need to be specially notified of an urgent window if I can't already see it.

And just a quick note not necessarily related to this PR:

Thank you, @IvanMalison!  When in #229 (which is awesome) you suggested deprecating the WorkspaceSwitcher widget, I was apprehensive because I thought the HUD widget was strictly about eye candy when I really like my simple text-based switcher.  But over the course of the afternoon I was able to get my exact switcher configuration working with the HUD widget, and got to eliminate some of my hacks in the process.  I love it!

screenshot_20171015_221318
(my pager showing active, visible, and urgent workspaces)

```haskell
workspaceHUDConfig :: WorkspaceHUDConfig
workspaceHUDConfig = defaultWorkspaceHUDConfig
    { widgetBuilder        = buildButtonController buildLabelController
    , minWSWidgetSize      = Nothing
    , labelSetter          = workspaceHUDLabelSetter
    , showWorkspaceFn      = hideEmpty
    , urgentWorkspaceState = True
    }

workspaceHUDLabelSetter :: Workspace -> HUDIO String
workspaceHUDLabelSetter workspace = do
    workspacesRef <- asks workspacesVar
    workspaces    <- lift $ readMVar workspacesRef
    let nonEmptyWorkspaces  = filter (\(_, w) -> workspaceState w /= Empty) $ M.toList workspaces
    let (lastWsIdx, _)      = last nonEmptyWorkspaces
    let wsIdx@(WSIdx wsNum) = workspaceIdx workspace
    let endMarker           = if wsIdx == lastWsIdx then "\x00a0" else ""
    let name                = escape $ (show $ wsNum + 1) ++ ": " ++ workspaceName workspace ++ endMarker
    return $ case workspaceState workspace of
        Active  -> workspaceDecorator "#282828" "#7cafc2" True $ name
        Visible -> workspaceDecorator "#d8d8d8" "#383838" False $ name
        Hidden  -> workspaceDecorator "#d8d8d8" "#282828" False $ name
        Urgent  -> workspaceDecorator "#282828" "#f7ca88" False $ name
        Empty   -> name
  where
    workspaceDecorator is the same decorator function I used
      in my WorkspaceSwitcher widget configuration
```

### DIFF
--- a/src/System/Taffybar/WorkspaceHUD.hs
+++ b/src/System/Taffybar/WorkspaceHUD.hs
@@ -22,6 +22,7 @@ module System.Taffybar.WorkspaceHUD (
   WorkspaceButtonController(..),
   WorkspaceContentsController(..),
   WorkspaceHUDConfig(..),
+  WorkspaceState(..),
   WorkspaceUnderlineController(..),
   WorkspaceWidgetController(..),
   IconController(..),
--- a/src/System/Taffybar/WorkspaceHUD.hs
+++ b/src/System/Taffybar/WorkspaceHUD.hs
@@ -57,7 +57,7 @@ import           Control.Monad.Reader
 import           Control.RateLimit
 import qualified Data.Char as S
 import qualified Data.Foldable as F
-import           Data.List (sortBy)
+import           Data.List (intersect, sortBy)
 import qualified Data.Map as M
 import           Data.Maybe
 import qualified Data.MultiMap as MM
@@ -338,10 +338,10 @@ buildWorkspaces _ = ask >>= \context -> liftX11Def M.empty $ do
   activeWindows <- readAsListOfWindow Nothing "_NET_ACTIVE_WINDOW"
   active:visible <- getVisibleWorkspaces
   let getWorkspaceState idx ws
-        | urgentWorkspaceState (hudConfig context) && not (null urgentWindows) =
-          Urgent
         | idx == active = Active
         | idx `elem` visible = Visible
+        | urgentWorkspaceState (hudConfig context) && not (null (ws `intersect` urgentWindows)) =
+          Urgent
         | null ws = Empty
         | otherwise = Hidden
   foldM


src/System/Taffybar/WorkspaceHUD.hs
```diff
@@ -337,10 +338,10 @@ buildWorkspaces _ = ask >>= \context -> liftX11Def M.empty $ do
   activeWindows <- readAsListOfWindow Nothing "_NET_ACTIVE_WINDOW"
   active:visible <- getVisibleWorkspaces
   let getWorkspaceState idx ws
-        | urgentWorkspaceState (hudConfig context) && not (null urgentWindows) =
-          Urgent
         | idx == active = Active
         | idx `elem` visible = Visible
+        | urgentWorkspaceState (hudConfig context) && not (null (ws `intersect` urgentWindows)) =
```
IvanMalison: so this makes it so that the "active" property of workspaces takes precedence over the urgent property. Now that I think about it, I guess that seems more correct.

I wonder if in the future we should make it possible to configure this, although I'm fine with it for now.

IvanMalison:
> I thought the HUD widget was strictly about eye candy when I really like my simple text-based switcher. But over the course of the afternoon I was able to get my exact switcher configuration working with the HUD widget, and got to eliminate some of my hacks in the process. I love it!

@iamjamestl 
That's awesome to hear. Though it does have built in support for a lot of different eye candy, WorkspaceHUD has always been about making the switcher as configurable as possible.

I think that with this sort of widget in particular, people seem to want really different sorts of looks and types of information, so this is essential for avoiding the problem of having someone write a new switcher every time one doesn't do exactly what they want out of the box.

Another thing to note (for those who aren't aware) is that the function `hudFromPagerConfig` actually builds a workspaceHUD widget that should look identical to the WSS widget that would be built from that config.

Anyway, I'm super happy to hear that it worked out for you

> (my pager showing active, visible, and urgent workspaces)

Looks awesome. I think it would be cool to have a collection of the different looks people get from taffybar. Probably wouldn't be hard at all to write a little snippet of code to automatically generate an image.

LGTM with this change!

src/System/Taffybar/WorkspaceHUD.hs
iamjamestl: My experience with the urgency setting before this change was that I'd be typing into vim, which likes to make a lot of beeps, and with the urxvt `urgentOnBell` option on, my active workspace to be marked urgent every few seconds.  That should probably be fixed by determining workspace urgency without considering active windows (something like `urgentWindows <- filterM isWindowUrgent (wins \\ activeWindows)`).  Then, I think it would be reasonable to want to be able to configure whether urgency takes precedence over active and/or visible states.  My opinion is that if you can see the window, you don't need a workspace notification about it, but I could understand how others would want a notification.  I will open an issue to discuss this further.

IvanMalison:
@travitch Merge? This looks pretty good to me.

PR #234, opened 2017-10-20
Add freedesktop menu widget

u11gh:
Here is some code for a menu widget that is filled as described by the
freedesktop (aka xdg) menu specification.  It is supposed show the
same entries as the "regular" Mate-, Xfce- or whatever-menu.

Create and use like other widgets: `let menu = menuWidgetNew (Just "mate")`, 
`startWidgets = [menu, ...]`

Implementation of the freedesktop standard is probably incomplete, but
I could imagine that there may be someone who might find it useful
anyway.


travitch:
Thanks for the well-commented commit and feature.  I'm happy to merge.  I didn't check closely, but my only question is: are any of the helpers dealing with XDG environment variables subsumed by the xdg-basedir package (https://hackage.haskell.org/package/xdg-basedir)?

u11gh:
I am very happy to see that you merged already.  Thanks!  :)

IIRC xdg-basedir did not provide all the xdg things I needed so I
decided not to use it at all (for sake of consistency).

Issue #240, opened 2017-11-12
freedesktop menu widget breaks on nonexistent XDG_CONFIG_DIRS items, empty menu prefix


benley:
built at 2647bc9071fb926d460e447ba1c333c06736b7d7

I'm running taffybar on NixOS, which sets various XDG_* variables somewhat differently than typical linux distros, and the menu widget is failing to load the system menus:

```
$ result/sw/bin/taffybar 
Launching custom binary /home/benley/.cache/taffybar/taffybar-linux-x86_64

Reading /home/benley/.nix-profile/etc/xdg/menus/gnome-applications.menu
taffybar-linux-x86_64: /home/benley/.nix-profile/etc/xdg/menus/gnome-applications.menu: openFile: does not exist (No such file or directory)
```

Relevant environment:
```
$ env|grep XDG_
XDG_VTNR=7
XDG_SESSION_ID=6
XDG_SESSION_TYPE=x11
XDG_DATA_DIRS=/run/opengl-driver/share:/run/opengl-driver-32/share:/etc/per-user-pkgs/benley/share:/home/benley/.nix-profile/share:/nix/var/nix/profiles/default/share:/run/current-system/sw/share
XDG_SESSION_DESKTOP=
XDG_SESSION_CLASS=user
XDG_SEAT_PATH=/org/freedesktop/DisplayManager/Seat0
XDG_CURRENT_DESKTOP=
XDG_SEAT=seat0
XDG_RUNTIME_DIR=/run/user/1000
XDG_SESSION_PATH=/org/freedesktop/DisplayManager/Session3
XDG_CONFIG_DIRS=/etc/per-user-pkgs/benley/etc/xdg:/home/benley/.nix-profile/etc/xdg:/nix/var/nix/profiles/default/etc/xdg:/run/current-system/sw/etc/xdg
```

It looks like XDG_CONFIG_DIRS is the one that affects this.  Of the items listed in that variable, only `/run/current-system/sw/etc/xdg` contains a `menus` directory, but `getXdgConfigDir` just checks for the existence of the `..../etc/xdg` directory.  If there is no `menus` subdirectory, taffybar crashes.

Also, `menuWidgetNew` does not work in the case where XDG_MENU_PREFIX is empty and only `applications.menu` exists.  It currently requires that _some_ `*-applications.menu` file exists in an xdg config dir, but that's not always going to work.  On my system:

```
$ ls /run/current-system/sw/etc/xdg/menus/
[...]

PR #237, opened 2017-10-25
Migrate to gtk3

IvanMalison:
So I got gtk3 working, mostly in the hopes of resolving #231, #235 and #228.
I was able to get the tray working, so #236 won't be necessary, although I did also have to make some changes to gtk-traymanager. I'll make a pull request there shortly. This means that taffybar will need a bump to gtk-traymanager to be cabal installable. Any help with testing with this would really be appreciated. I think there are still some widgets I haven't totally checked out so anyone who can help (@iamjamestl @teleshoes @sakshamsharma @escherdragon @frioux @u11gh @shwsun @travitch). I'd also love help seeing if #231 might actually be fixed by this.

Things to note about this change:
- gtk completely changes theming support to be css based, so you'll need to migrate your rc file to css to do this. The good news is that css is way more flexible and powerful. It also offers way more features (Here's a sample with some opacity and rounded corners: https://imgur.com/a/Amdil)
- You will need [my updates](https://github.com/travitch/gtk-traymanager/pull/8) to gtk-traymanager to get things to compile. In stack you can set this up with
``` yaml
- location:
    git: git@github.com:IvanMalison/gtk-traymanager.git
    commit: 39cddcf77f3e4a8ecbc4d82894457fd61c1faff1
  extra-dep: true
```

fixes #121, fixes #110, fixes #152, fixes #145, fixes #236
Contains all of #126 though we haven't heard from (@nisstyre56 and @f1u77y who I think have stopped using taffybar)

PR #242, opened 2017-11-16
Handle cases where a config dir exists but the menu file in it does not

benley:
Don't give up after checking the first one; continue down the list of
possible paths until we find a menu file that exists.

More fixes for #240

### DIFF
--- a/src/System/Taffybar/Menu/XdgMenu.hs
+++ b/src/System/Taffybar/Menu/XdgMenu.hs
@@ -27,20 +27,25 @@ module System.Taffybar.Menu.XdgMenu (
   getPreferredLanguages)
 where
 
-import GHC.IO.Encoding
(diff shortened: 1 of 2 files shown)


benley:
@u11gh here's my attempt at further fixing #240 and #241.  I'm still pretty new to haskell so hopefully the change makes sense.

IvanMalison:
You need to import <$> to make this compile in 7.8, although I have already done that in the relevant file with my changes...

benley:
I'll try to redo this atop your changes now that I have a working version :-)

benley:
OK, please take another look.  I'm not aware of a good way to properly cherrypick your cleanups from #237 using git to preserve authorship, but I did at least port the changes over in a standalone commit so the diffs are coherent.

PR #248, opened 2018-01-29
Add GHC 8.2.2 to the Travis build matrix.

LeifW:


LeifW:
The 'Test quick start" part failed - I wouldn't think adding a new configuration to the build matrix would affect that.

IvanMalison:
The build error seems to have to do with the dyre dependency

IvanMalison:
#249 Should fix the build.

PR #251, opened 2018-02-16
Add HostName widget

### LINKED ISSUE #250: widget to show the current hostname
It would be nice if there was a stock widget to show the current hostname.  This can easily be added with something like `commandRunnerNew (60 * 60) "uname" ["--nodename"] "<hostname error>" "red"`, but it would be more convenient if there was a stock widget to do it.

If I sent a PR adding this, would it be accepted?

cdepillabout:
This PR adds a HostName widget.  This shows the current machine hostname.

It is relatively simple, but I have a question about it.  Should it poll for a new hostname?  That is to say, should it be based on a `PollingLabelWidget`?  It don't know how frequently it is for people to change their hostname, but for me, I never change my hostname, so I don't need any sort of polling functionality.

If this hostname widget should poll for an updated hostname, then the `HostNameConfig` data type should probably have a field to specify the polling interval.

It if should not poll for an updated hostname, then a `PollingLabelWidget` should probably not be used.  Maybe I should create a new non-polling `LabelWidget`?  That way it could be reused by other widgets.

Closes #250


src/System/Taffybar/HostName.hs
```diff
@@ -0,0 +1,51 @@
...
+import Data.Monoid ((<>))
+import Graphics.UI.Gtk (Widget, toWidget, widgetShowAll)
+import Network.HostName (getHostName)
+
+import System.Taffybar.Widgets.PollingLabel (pollingLabelNew)
+
+-- | The configuration for the hostname label.
+data HostNameConfig =
+  HostNameConfig
+    { hostNameBackgroundColor :: Maybe String -- ^ Background color. If
```
IvanMalison: can you put this comment on its own line so that the columns aren't so short.

Use -- | if it makes you happy

src/System/Taffybar/HostName.hs
```diff
@@ -0,0 +1,51 @@
...
+
+getHostNameString :: HostNameConfig -> IO String
+getHostNameString cfg = do
+  hostName <- getHostName
+  let bgColorFmt =
+        case hostNameBackgroundColor cfg of
+          Nothing -> ""
+          Just bgColor -> "bgcolor='" <> bgColor <> "' "
+  return $
+    "<span fgcolor='" <>
```
IvanMalison: Would be much better to use theming support imo. I know that things are done this way in a few places in taffybar atm, but it would be nice to have things done in a more uniform way in the future (especially once the migration to gtk3 is done.

src/System/Taffybar/HostName.hs
cdepillabout: No problem.  I'll fix this.

src/System/Taffybar/HostName.hs
cdepillabout: You're right.  Using the theming support would be better.

cdepillabout:
@IvanMalison Thanks for the review.  I've pushed commit 448f623 that addresses your comments.

What do you think of how I've added theming support?  Is there a better way to do it?

IvanMalison:
@cdepillabout I meant to add theming support as in

https://github.com/travitch/taffybar/blob/2647bc9071fb926d460e447ba1c333c06736b7d7/src/System/Taffybar/WorkspaceHUD.hs#L844

In the new gtk3 version this will be css theming instead of gtks weird stylesheets as in:

https://github.com/travitch/taffybar/blob/92d1df8a2035142b3e05620db19aff30ef3d566d/src/System/Taffybar/WorkspaceHUD.hs#L153

cdepillabout:
@IvanMalison Is there an example of using GTK's theming support in another widget currently in the code?  I'm afraid I don't understand how `Gtk.widgetSetName` is related to theming support.  Did you maybe mean to link to another part of the code?

In my original code, I am just using `pollingLabelNew`.  The `pollingLabelNew` function doesn't seem to take any arguments related to GTK theming.  Am I missing something here?

cdepillabout:
I should add that I am not very familiar with GTK.  Is theming support some sort of basic functionality of GTK?

Since I am just using pollingLabelNew, most of the code in my original PR doesn't do that much GTK stuff.  A simple example that shows how to set the foreground color for a label using the theming functionality of GTK would probably help me understand what you're talking about.

IvanMalison:
widgetSetName allows you to target the widget that has been named with your taffybar.rc (taffybar.css in gtk3) file. It works the way that css does in that you specify the theme/appearance properties separately.

You can take a look at the taffybar.rc file in the root of the repository

cdepillabout:
@IvanMalison Is there some sort of documentation about taffybar.rc?  There is no mention of it in the README.  There is a one-sentence explanation of it in the System.Taffybar documentation, but the default theme does not exist on my machine.  (Maybe because I'm compiling with stack instead of cabal?)  Also, the link in the documentation doesn't work.

As an end user, I don't want to have two separate configuration files that I must update to configure taffybar.  I would much rather specify everything in `taffybar.hs`.

If I use `widgetSetName`, is there a way to override the color specified in `taffybar.rc` and just use the color specified in `HostNameConfig`?  None of the other widgets I've looked at seemed to be using `widgetSetName`.

IvanMalison:
> @IvanMalison Is there some sort of documentation about taffybar.rc?

Nothing that is taffybar specific. Here is something on gtkrc files in general:

https://www.gtk.org/tutorial1.2/gtk_tut-21.html

> As an end user, I don't want to have two separate configuration files that I must update to configure taffybar. I would much rather specify everything in taffybar.hs.

I WOULD be inclined to agree with this sentiment, but unfortunately it seems that certain types of appearance properties including rounded edges and transitions are actually not possible without the use of rc files(actually css files in gtk3).

> If I use widgetSetName, is there a way to override the color specified in taffybar.rc and just use the color specified in HostNameConfig?

I believe that if you set the color of a widget explicitly in code that will take precedence over any appearance stuff that is set through css/rc I'm not 100% sure on that one though.

> None of the other widgets I've looked at seemed to be using widgetSetName.

WorkspaceHUD, WorkspaceSwitcher and WindowSwitcher all use it. Many widgets in taffybar don't really need to be explicitly themed as long as they just use the default top level taffybar themes (rc/css files use a type of selector that can be applied to all children), but it is true that there are a few spots where theming works differently.

While I can't say that I love having to edit a separate file to control certain things about taffybar's appearance, we have to do things that way to achieve certain visual effects (as far as i can tell). As such, I think that its better that we have all of the color/theming stuff in one place as much as possible. There are certain advantages to doing things with css (selectors /classes etc.), especially if you want to share theming across different widgets.

IvanMalison:
You might want to try to work with #237 as a base for your changes since that is likely to get merged soon, and it is possible that some of the stuff you are using here will need to be slightly different in gtk3.

IvanMalison:
#237 is merged, and now you seem to have some conflicts.

Now instead of widgetSetName, you'll want to use something like:

https://github.com/travitch/taffybar/blob/b2e9577b3de272f56f9c89dfb01c3d3d9107cf55/src/System/Taffybar/WorkspaceHUD.hs#L143

to add a css class.

cdepillabout:
Unfortunately work has gotten busy for me, so I won't have any time to continue with this PR.  As I mentioned on #250, I can use `commandRunnerNew` for this, so it is not critical that it exists.

However, I have `Allow edits from maintainers` checked, so if you would like to continue working on it, please feel free.

Otherwise feel free to close this PR.

IvanMalison:
You can very easily use:

`commandRunnerNew 60.0 "hostname" [] "failed"`

PR #252, opened 2018-02-22
Add auto format for netmonitor

TeXitoi:
Will output, with precision 3:
 - 0B/s
 - 1023B/s
 - 1.00KiB/s
 - 42.3KiB/s
 - 123MiB/s
 - 1002GiB/s

I've done an extra commit to change the default to use it, but if you prefer to let the old version, I can remove this commit.


IvanMalison:
Cool change. I'm going to try to merge in on top of #237 to make sure all is well on that front.

PR #289, opened 2018-04-21
Expose pollingBarNew sans threadDelay

LeifW:
For displaying e.g. screen brightness, I figured using poll(2) or select
or the like would be more efficient and give better response times than
simply reading the file at a periodic interval.

The bulk of that function looks the same as the existing pollingBarNew,
just without the threadDelay - the existing function can be implemented
in terms of the "new" one just by tacking on a threadDelay to the
action.

For my use with "polling" the value out of /sys/class/backlight/intel_backlight/actual_brightness, my invocation of this looks something like `verticalBarFromCallback cfg $ readNum path <* watchFile path`, where `watchFile is a wrapper for https://github.com/quixoftic/hpio/blob/master/src/System/GPIO/Linux/Sysfs/IO.hs#L132, so it will just block that thread until that file changes.

Not sure what good naming for this stuff is.

### DIFF
--- a/src/System/Taffybar/Widgets/PollingBar.hs
+++ b/src/System/Taffybar/Widgets/PollingBar.hs
@@ -7,27 +7,27 @@ module System.Taffybar.Widgets.PollingBar (
   BarDirection(..),
   -- * Constructors and accessors
   pollingBarNew,
+  verticalBarFromCallback,
   defaultBarConfig
   ) where
 
 import Control.Concurrent
-import qualified Control.Exception.Enclosed as E
+import Control.Exception.Enclosed ( tryAny )
 import Control.Monad ( forever )
 import Graphics.UI.Gtk
+import System.Taffybar.Widgets.Util ( backgroundLoop, drawOn )
 
 import System.Taffybar.Widgets.VerticalBar
 
-pollingBarNew :: BarConfig -> Double -> IO Double -> IO Widget
-pollingBarNew cfg pollSeconds action = do
+verticalBarFromCallback :: BarConfig -> IO Double -> IO Widget
+verticalBarFromCallback cfg action = do
   (drawArea, h) <- verticalBarNew cfg
+  drawOn drawArea $
+    backgroundLoop $ do
+      esample <- tryAny action
+      traverse (verticalBarSetPercent h) esample
 
-  _ <- on drawArea realize $ do
-    _ <- forkIO $ forever $ do
-      esample <- E.tryAny action
-      case esample of
-        Left _ -> return ()
-        Right sample -> verticalBarSetPercent h sample
-      threadDelay $ floor (pollSeconds * 1000000)
-    return ()
-
-  return drawArea
+pollingBarNew :: BarConfig -> Double -> IO Double -> IO Widget
+pollingBarNew cfg pollSeconds action =
+  verticalBarFromCallback cfg $ action <* delay
+  where delay = threadDelay $ floor (pollSeconds * 1000000)
--- a/src/System/Taffybar/Widgets/Util.hs
+++ b/src/System/Taffybar/Widgets/Util.hs
@@ -14,11 +14,12 @@
 
 module System.Taffybar.Widgets.Util where
 
-import Control.Monad
+import Control.Monad ( when, forever, void )
 import Control.Monad.IO.Class
+import Data.Functor ( ($>) )
 import Data.Tuple.Sequence
+import Control.Concurrent ( forkIO )
 import Graphics.UI.Gtk
-import Prelude
 import Text.Printf
 
 -- | Execute the given action as a response to any of the given types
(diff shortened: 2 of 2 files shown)


src/System/Taffybar/Widgets/PollingBar.hs
```diff
@@ -7,27 +7,27 @@ module System.Taffybar.Widgets.PollingBar (
...
-      case esample of
-        Left _ -> return ()
-        Right sample -> verticalBarSetPercent h sample
-      threadDelay $ floor (pollSeconds * 1000000)
-    return ()
-
-  return drawArea
+pollingBarNew :: BarConfig -> Double -> IO Double -> IO Widget
+pollingBarNew cfg pollSeconds action =
+  verticalBarFromCallback cfg $ action <* delay
```
LeifW: Actually, this introduces a bit of a change - because the threadDelay is now inside of the action, it needs to complete before verticalBarSetPercent is called. The net effect is that your bar can take up to pollSeconds to initally draw when taffybar is launched, but other than that I don't think it'd have any effect. I could try thinking of something else if that's not acceptable.
Also, I don't think `action <* delay` vs `delay *> action` really matters here.

LeifW: Perhaps splitting up the `action` that `verticalBarFromCallback` takes into two IO values - one of which blocks, and the other of which produces a value, so it can immediately paint the initial value, and then go into its blocking / polling loop - rather than having them combined into one with `*>` or `do` or whatever.

LeifW: I tried a refactor like that locally:
```haskell
do
  action >>= verticalBarSetPercent h
  wait
```
And it doesn't seem to have any effect - it still waits until the first time I adjust the brightness to display a value on the brightness graph.

IvanMalison: I know that you didn't introduce this, but one thing that I really don't like about this approach is that it doesn't account for the fact that the action itself could take a (possibly variable) amount of time.

This means that the timing between runs might not really actually be very accurate in practice. If our action takes a non-trivial amount of time, say .5 times the interval, then it could end up getting updated far less than the user expects.

LeifW: Makes sense. Not many of the "get data" actions I have take a non-trivial amount of time, other than the one making an HTTP request to the weather service, and I'm not really worried about the accuracy of the update interval there (as long as it gets updated, period, at some point), but it does seem like something to keep in mind. I did see a few scheduling libs out there.

IvanMalison:
@LeifW I'll take a look at this approach. I believe that the uPower DBus interface (for which we already have a generated client provides some backlight support)

LeifW:
Or maybe udev? I think I've seen udev used with epoll. Here's I'm using the udev package to read from e.g. `/sys/bus/iio/devices/iio:device1/in_accel_x_raw`
https://github.com/LeifW/rotate-screen/blob/master/UdevAccelerometer.hs#L69
I could see if I could read backlight values that way.

LeifW:
Turns out these values can also be read with udev - if that would be useful.
An enumerate with `addMatchSubsystem e "backlight"` gives me a device path of `/sys/devices/pci0000:00/0000:00:02.0/drm/card0/card0-eDP-1/intel_backlight`, and then I can call `getSysattrValue dev "actual_brightness"` on that device to get the brightness.

LeifW:
If I run this example program included with the udev package: https://github.com/pxqr/udev/blob/master/examples/monitor.hs
Every time I hit a change brightness button, it spits out:
```
Just "backlight"
Nothing
"/sys/devices/pci0000:00/0000:00:02.0/drm/card0/card0-eDP-1/intel_backlight"
"intel_backlight"
Nothing
Nothing
Just Remove
device number: Devnum {major = 0, minor = 0}
```
It also uses the `select'` function from the `select` package - a wrapper for `select(2)`, to poll.

So it seems udev + select or epoll or whatever would be an option for this case, if desired.

IvanMalison:
@LeifW Cool that looks like a great way to do it. PR welcome.

One pattern that I've been trying to use is to separate the concern of monitoring from the actual widget display. This means creating an information module that exposes a channel that can be shared through the extensible state mechanism in context, and then having each widget that wants access to updates retrieve that channel.

For examples of this pattern, check out

Battery: https://github.com/taffybar/taffybar/blob/0d584e4efcb7f2802ce23acc90309bd77e113f12/src/System/Taffybar/Information/Battery.hs#L153

Network:
https://github.com/taffybar/taffybar/blob/0d584e4efcb7f2802ce23acc90309bd77e113f12/src/System/Taffybar/Hooks.hs#L22
https://github.com/taffybar/taffybar/blob/0d584e4efcb7f2802ce23acc90309bd77e113f12/src/System/Taffybar/Information/Network.hs#L99

Once you set up this pattern, you can write the widget really easily using ChannelWidget
https://github.com/taffybar/taffybar/blob/0d584e4efcb7f2802ce23acc90309bd77e113f12/src/System/Taffybar/Widget/Generic/ChannelWidget.hs#L8

For an example of this, check out 
https://github.com/taffybar/taffybar/blob/0d584e4efcb7f2802ce23acc90309bd77e113f12/src/System/Taffybar/Widget/Text/NetworkMonitor.hs#L62

PR #312, opened 2018-05-08
Fix a battery search.

iloskutov:
It fixes a battery search according to UPower object path generation.

### DIFF
--- a/src/System/Taffybar/Information/Battery.hs
+++ b/src/System/Taffybar/Information/Battery.hs
@@ -21,7 +21,7 @@ import Data.Word
 import Data.Int
 import DBus
 import DBus.Client
-import Data.List ( isInfixOf )
+import Data.List ( isPrefixOf )
 import Data.Text ( Text )
 import qualified Data.Text as T
 import Safe ( atMay )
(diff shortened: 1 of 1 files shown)


IvanMalison:
@iloskutov Hmm, the battery widget DOES seem to work for some people right now, so I think maybe we should make this check for both prefixes. Can you link the justification for this change?

iloskutov:
It works on one of my laptops but didn't work on other. On first 'upower -e' shows:
/org/freedesktop/UPower/devices/line_power_AC                                                                                                                  
/org/freedesktop/UPower/devices/battery_BAT0                                                                                                                   
/org/freedesktop/UPower/devices/battery_BAT1                                                                                                                   
/org/freedesktop/UPower/devices/DisplayDevice 
On second one:
/org/freedesktop/UPower/devices/battery_max170xx_battery
/org/freedesktop/UPower/devices/line_power_fusb302_typec_source
/org/freedesktop/UPower/devices/line_power_bq24190_charger
/org/freedesktop/UPower/devices/DisplayDevice
I looked at sources of UPower and found that the object names are generated by function up_device_compute_object_path -  [https://github.com/freedesktop/upower/blob/master/src/up-device.c#L347](https://github.com/freedesktop/upower/blob/master/src/up-device.c#L347)
For battery the type will be "battery", and basename of native_path gets from corresponding device in /sys/class/power_supply/.
I think filtering by "battery_" will be more correct.

IvanMalison:
@iloskutov I see, didnt realize thats how the naming was working. In taht case, lets make it isPrefixOF instead of isInfixOf

iloskutov:
@IvanMalison, is it better?

PR #328, opened 2018-05-13
PollingLabel shouldnt strip whitespace

teleshoes:
IMO, formatting shouldnt be the responsibility of the generic widget, for the same reason that HTML entities are not stripped by this widget `Proper input sanitization is up to the caller`

there is, of course, a practical reason not to strip whitespace as well; stripping the whitespace breaks my fixed-width monospace labels


IvanMalison:
@teleshoes okay, alot of commands will output trailing newlines, which will make things confusing for new users, but I agree in principle, so I'll merge this.

teleshoes:
yay! thanks!

PR #347, opened 2018-05-18
Brightness widget; looked up by udev and polled by udev monitor

### LINKED ISSUE #321: Backlight widget
Add a backlight widget to monitor screen lighting

LeifW:
On my system, looking up a udev device of subsystem "backlight" gives me the device corresponding to `/sys/class/backlight/intel_backlight`
Until the next udev version is published, we're not filtering on events in udev, just inspecting all events received to see if they're from the "backlight" subsystem.

Perhaps ChannelWidget can / should be leveraged - maybe the stuff in PollingBar could be consolidated with that.
The stuff I added is missing the `void $ onWidgetUnrealize widget $ killThread processingThreadId` at the end, for instance. And also stuff to kill the data source thread - in this case, the one waiting on the udev monitor.

Fixes #321

### DIFF
--- a/src/System/Taffybar/Information/Brightness.hs
+++ b/src/System/Taffybar/Information/Brightness.hs
@@ -1,0 +1,51 @@
+{-# LANGUAGE OverloadedStrings #-}
+module System.Taffybar.Information.Brightness where
+
+import System.UDev.Context
+import System.UDev.Device
+import System.UDev.Monitor
+
+import System.UDev.Enumerate
+import System.UDev.List
+
+import Control.Concurrent (Chan, newChan, writeChan, threadWaitRead)
+
+import Control.Monad (when)
+
+import Data.ByteString (ByteString())
+
+import System.Taffybar.Util (backgroundLoop)
+import System.Taffybar.Util.ReadByteString (parseInt)
+
+--Only returns the first match
+scanForDeviceBySubsystem :: UDev -> Subsystem -> IO Device
+scanForDeviceBySubsystem udev subSysName = do
+  e <- newEnumerate udev
+  addMatchSubsystem e subSysName
+  scanDevices e 
+  Just ls <- getListEntry e
+  path <- getName ls
+  newFromSysPath udev path
+  
+getInt :: ByteString -> Device -> IO Int
+getInt name dev = parseInt <$> getSysattrValue dev name
+
+readBrightnessValues :: IO (Chan Double)
+readBrightnessValues = withUDev $ \udev -> do
+  max_brightness <- scanForDeviceBySubsystem udev "backlight" >>= getInt "max_brightness"
+  monitor <- newFromNetlink udev UDevId
+  -- Uncomment this when next version of udev is released:
+  -- filterAddMatchSubsystemDevtype monitor "backlight" Nothing
+  enableReceiving monitor
+  fd <- getFd monitor
+  chan <- newChan
+  backgroundLoop $ do
+    threadWaitRead fd
+    dev <- receiveDevice monitor
+    -- replace this guard with above filterAddMatchSubsystemDevtype when next udev is released
+    when (getSubsystem dev == Just "backlight") $ do
+      actual_brightness <- getInt "actual_brightness" dev
+      let percent = fromIntegral actual_brightness / fromIntegral max_brightness
+      writeChan chan percent
+  return chan
+
--- a/src/System/Taffybar/Information/Brightness.hs
+++ b/src/System/Taffybar/Information/Brightness.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
-module System.Taffybar.Information.Brightness where
+module System.Taffybar.Information.Brightness (readBrightnessValues) where
 
 import System.UDev.Context
 import System.UDev.Device
--- a/src/System/Taffybar/Information/Brightness.hs
+++ b/src/System/Taffybar/Information/Brightness.hs
@@ -23,7 +23,7 @@ scanForDeviceBySubsystem udev subSysName = do
   e <- newEnumerate udev
   addMatchSubsystem e subSysName
   scanDevices e 
-  Just ls <- getListEntry e
+  ls <- getListEntry e -- This becomes a Maybe in the next udev release
   path <- getName ls
   newFromSysPath udev path
   
--- a/src/System/Taffybar/Util.hs
+++ b/src/System/Taffybar/Util.hs
@@ -82,6 +82,15 @@ foreverWithDelay :: RealFrac a1 => a1 -> IO a -> IO ThreadId
 foreverWithDelay delay action =
   forkIO $ forever $ action >> threadDelay (floor $ delay * 1000000)
 
+backgroundLoop :: IO a -> IO ()
+backgroundLoop = void . forkIO . forever
+
+loopActionToChan :: IO a -> IO (Chan a)
+loopActionToChan action = do
+  chan <- newChan
+  backgroundLoop $ action >>= writeChan chan
+  return chan
+
 liftActionTaker
   :: (Monad m)
   => ((a -> m a) -> m b) -> (a -> ReaderT c m a) -> ReaderT c m b
--- a/src/System/Taffybar/Util.hs
+++ b/src/System/Taffybar/Util.hs
@@ -85,6 +85,7 @@ foreverWithDelay delay action =
 backgroundLoop :: IO a -> IO ()
 backgroundLoop = void . forkIO . forever
 
+-- This could take an IO (Maybe a) instead of the action doesn't return a new value on every exexecution?
 loopActionToChan :: IO a -> IO (Chan a)
 loopActionToChan action = do
   chan <- newChan
--- a/src/System/Taffybar/Util/ReadByteString.hs
+++ b/src/System/Taffybar/Util/ReadByteString.hs
@@ -1,0 +1,14 @@
+{-# LANGUAGE OverloadedStrings #-}
+module System.Taffybar.Util.ReadByteString (parseInt) where
+
+import Data.ByteString (ByteString())
+import Data.Text (Text())
+import Data.Text.Encoding (decodeUtf8)
+import Data.Text.Read (decimal)
+
+--partial
+fromReader :: Either String (a, Text) -> a
+fromReader (Right (n, "")) = n
+
+parseInt :: ByteString -> Int
+parseInt = fromReader . decimal . decodeUtf8
--- a/src/System/Taffybar/Widget/Brightness.hs
+++ b/src/System/Taffybar/Widget/Brightness.hs
@@ -1,11 +1,13 @@
 module System.Taffybar.Widget.Brightness (brightnessBarNew) where
 
+import Control.Concurrent.Chan (dupChan)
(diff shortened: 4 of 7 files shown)


LeifW:
Adds a dep to udev, which seems old and is a thin wrapper around the C apis of libudev. But it may also be useful for other things (the only other evens I've seen from running `udevadmn monitor` are the ones from me plugging or unplugging the AC adapter).
The libudev API is better documented at https://pyudev.readthedocs.io/en/latest/

taffybar.cabal
```diff
@@ -83,6 +84,7 @@ library
                , transformers >= 0.3.0.0
                , transformers-base >= 0.4
                , tuple >= 0.3.0.2
+               , udev == 0.1.0.0
```
IvanMalison: @LeifW you need to add udev to the stack extra deps. Make sure to get the stack.yaml.example file too

src/System/Taffybar/Information/Brightness.hs
```diff
@@ -0,0 +1,51 @@
+{-# LANGUAGE OverloadedStrings #-}
+module System.Taffybar.Information.Brightness (readBrightnessValues) where
+
+import System.UDev.Context
+import System.UDev.Device
+import System.UDev.Monitor
+
```
IvanMalison: Can we get rid of these random spaces between imports. I don't really like grouping them in any way. They are always just sorted lexically.

src/System/Taffybar/Information/Brightness.hs
```diff
@@ -0,0 +1,51 @@
...
+import System.UDev.Enumerate
+import System.UDev.List
+
+import Control.Concurrent (Chan, newChan, writeChan, threadWaitRead)
```
IvanMalison: I try to avoid any specific name imports. If you are worried about name collisions I prefer a qualified import e.g.
``` haskell
import qualified Control.Concurrent as C
```

src/System/Taffybar/Information/Brightness.hs
```diff
@@ -0,0 +1,51 @@
...
+
+import Control.Monad (when)
+
+import Data.ByteString (ByteString())
+
```
IvanMalison: as mentioned above, no newlines between imports

src/System/Taffybar/Information/Brightness.hs
```diff
@@ -0,0 +1,51 @@
...
+import System.Taffybar.Util (backgroundLoop)
+import System.Taffybar.Util.ReadByteString (parseInt)
+
+--Only returns the first match
+scanForDeviceBySubsystem :: UDev -> Subsystem -> IO Device
+scanForDeviceBySubsystem udev subSysName = do
+  e <- newEnumerate udev
```
IvanMalison: What is e? I can't tell from reading this because there is no type annotation. Generally I'd prefer slightly longer somewhat descriptive names

src/System/Taffybar/Information/Brightness.hs
```diff
@@ -0,0 +1,51 @@
...
+  newFromSysPath udev path
+  
+getInt :: ByteString -> Device -> IO Int
+getInt name dev = parseInt <$> getSysattrValue dev name
+
+readBrightnessValues :: IO (Chan Double)
+readBrightnessValues = withUDev $ \udev -> do
+  max_brightness <- scanForDeviceBySubsystem udev "backlight" >>= getInt "max_brightness"
+  monitor <- newFromNetlink udev UDevId
+  -- Uncomment this when next version of udev is released:
```
IvanMalison: Add a TODO or XXX here so this is easy to find.

src/System/Taffybar/Util.hs
```diff
@@ -82,6 +82,16 @@ foreverWithDelay :: RealFrac a1 => a1 -> IO a -> IO ThreadId
 foreverWithDelay delay action =
   forkIO $ forever $ action >> threadDelay (floor $ delay * 1000000)
 
+backgroundLoop :: IO a -> IO ()
+backgroundLoop = void . forkIO . forever
+
+-- This could take an IO (Maybe a) instead of the action doesn't return a new value on every exexecution?
```
IvanMalison: Sure, do that and remove the comment.

src/System/Taffybar/Information/Brightness.hs
```diff
@@ -0,0 +1,51 @@
...
+  -- filterAddMatchSubsystemDevtype monitor "backlight" Nothing
+  enableReceiving monitor
+  fd <- getFd monitor
+  chan <- newChan
+  backgroundLoop $ do
+    threadWaitRead fd
+    dev <- receiveDevice monitor
+    -- replace this guard with above filterAddMatchSubsystemDevtype when next udev is released
```
IvanMalison: add TODO

src/System/Taffybar/Util/ReadByteString.hs
```diff
@@ -0,0 +1,14 @@
+{-# LANGUAGE OverloadedStrings #-}
```
IvanMalison: I'm not convinced this needs its own module

src/System/Taffybar/Widget/Brightness.hs
```diff
@@ -0,0 +1,13 @@
+module System.Taffybar.Widget.Brightness (brightnessBarNew) where
```
IvanMalison: I've started adding imports for all widgets in System.Taffybar.Widget, so we should probably do that with this.

src/System/Taffybar/Widget/Brightness.hs
```diff
@@ -0,0 +1,13 @@
...
+
+import Control.Concurrent.Chan (dupChan)
+import Graphics.UI.Gtk (Widget)
+import System.Taffybar.Widget.Generic.VerticalBar (BarConfig)
+import System.Taffybar.Widget.Generic.PollingBar (verticalBarFromChannel)
+import System.Taffybar.Information.Brightness (readBrightnessValues)
+
+brightnessBarNew :: BarConfig -> IO Widget
```
IvanMalison: Yeah, this should just use ChannelWidget.

Also, this starts a new poller every time the widget is created. This means that if someone were to toggle taffybar repeatedly, you would start accumulating channels and threads that are polling this file.

I actually toggle taffybar on one of my monitors quite frequently (on the order of 10 times in the average lifecycle of a taffybar process). If i were to add this widget to my config, I would accumulate a whole bunch of these threads (and they would all be writing to an unused unbounded channel).

There are two ways to fix this. 
- One is to kill the channel/thread whenever the widget is unrealized.
- The better way to handle it is to use the pattern that I've been using for all of my channel widgets that leverages the state mechanism of Taffybar's Context reader to share a single channel for all widgets that want access to the information (This is the purpose of the dupChan btw. it makes the channel a broadcast channel)

You can look at
https://github.com/taffybar/taffybar/blob/a4d757c1509eb44526921c0d90afa157a23db1a7/src/System/Taffybar/Hooks.hs#L22
and
https://github.com/taffybar/taffybar/blob/a4d757c1509eb44526921c0d90afa157a23db1a7/src/System/Taffybar/Information/Battery.hs#L151

for examples of this pattern. Basically, the idea is to use getStateDefault to get the singleton channel instance.

IvanMalison: All widgets need to either be in TaffyIO or MonadIO m.

I don't think you would be able to add this to any of start, end or center widgets as is, right?

IvanMalison: It just occured to me that even with the approach that I currently use, there can be a space leak when there are no readers of the channel. We might want to look at using
https://github.com/merijn/broadcast-chan to handle this.

LeifW: I was wondering about the Chan thing - would something like MVar work? Reasoning being, you only care about the current value in the display. If for some reason you've missed a few updates, you don't care about the intermediate / historic values - just the current one.

IvanMalison: MVar doesn't really give you wakeup behavior. You can just empty the channel whenever you read it. Also when you dupchan you only receive values after the duplication, so unless you are too slow, this isn't going to happen.

LeifW: Oh yeah, takeMVar would only allow one thread to read the value. What about Control.Concurrent.Broadcast, with `listen` and `signal`? Allows multiple consumers to block on a single value, and all get it when a value is published.

IvanMalison:
@LeifW I'm going to be doing a release on Monday at around 12pm, and I'd like to try to sneak this in. Do you have time to do the fixes I mentioned or do you want me to pick it up?

LeifW:
I'll have some free time Sunday night. I can a lot of what you brought up. I still need to get up to speed with how monitors are spawned and how the Widgets consume them, though - would you be able to take it from there? I'm primarily familiar with the udev stuff in `System.Taffybar.Information.Brightness` that reads the values.

LeifW:
Judging by https://github.com/taffybar/taffybar/commit/bcddb9da16301eca023d1f17d21378c4f55c4042 I'm guessing this will go in a separate taffybar-linux-extras package - is that started, yet?

IvanMalison:
Does this add a non-haskell dependency?

LeifW:
The udev package, yes, which is bindings to the C lib libudev. Though, I'm having second thoughts on using that lib, given that my PR to allow it to build on GHC 8.4 has just been sitting there, and it was last published to hackage in 2018. Perhaps I could offer to take over maintainership? Another option would be just to read the values out of sysfs using regular file API - `/sys/class/backlight`. Can still use epoll or poll or whatever in there.

IvanMalison:
@LeifW Yea, In that case, it should go in taffybar-extra.

I just want to avoid adding extra difficulty to getting taffybar running, since it already has so many compile time dependencies. If there is some way we can make it runtime only it can stay here.

IvanMalison:
> Perhaps I could offer to take over maintainership?

Of udev, yeah I think thats probably a good idea. The author hasn't responded?  I think theres a way to get in touch with hackage people in this sort of scenario.

IvanMalison:
@LeifW I'll make the taffybar-extras module today

IvanMalison:
@LeifW What became of this. I actually want this for my new laptop.

IvanMalison:
@LeifW Any chance this could be done in the next week or 2 so that it can make the next release? What do I need to do if I pick it up?

taffybar.cabal
LeifW: This will restrict taffybar to Linux-only, as this is a binding to libudev. Did we want to make e.g. a separate taffybar-extras or taffybar-linux for this sort of thing, or hide that extra dep behind a configure flag and only add it conditionally (like xmonad-extras does for its extra deps?)

IvanMalison: I'm fine with either

src/System/Taffybar/Information/Brightness.hs
LeifW: It's essentially a C pointer, I guess: http://hackage.haskell.org/package/udev-0.1.0.0/docs/System-UDev-Enumerate.html#t:Enumerate
Kind of want to make reader-monad type things for these things to run in (all the following lines require that `e`)? Something like that could be upstreamed in the udev lib.  Though maybe I should go with the straightforward explicit stuff for now.

PR #364, opened 2018-05-30
Remove the rest of references to Volume

LeifW:


anarcat:
what happened here? was volume control introduced in 1.0 removed from later versions?

IvanMalison:
Volume control was removed because it required a libarary that would only install with linux only dependencies.

IvanMalison:
I think that pulseaudio has a dbus interface. We could write something that talks to that that wouldnt require a library that we could put in taffybar.

https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/Clients/DBus/

I've filed #428 for this.

LeifW:
We might want to have like a taffybar-linux-extras package you can install
on top of taffybar for things like that - there's other linux libs we can
use, like iwlib for showing wifi status.

PR #366, opened 2018-06-04
Fix example configuration

ARR8:
Using the quick start script, I was getting errors and warnings when it came to the last step. Eventually I found the module where the tray function was defined to fix it. Hopefully this will save other people some time.


IvanMalison:
@ARR8 I didn't realize that function wasn't exported from widgets from that change 9c53e85 makes it so that there is no need for this import.

Thanks for pointing this out though!

IvanMalison:
ahh.

PR #375, opened 2018-06-10
Fix#371 part2

### LINKED ISSUE #371: XDG Menu: Does not read /etc/xdg unless specified
**Describe the bug**
Unlike the description in the comments of the source file, the XDG menu widget seems to only search in /etc/xdg if it is manually specified in XDG_CONFIG_DIRS. Otherwise, there is an error that .config/menus/PREFIX-applications.menu does not exist, but it does not look further.

**To Reproduce**
Unset the XDG_CONFIG_DIRS environment variable and launch taffybar with the XDG menu.

**Expected behavior**
Following the XDG spec, check /etc/xdg even if nothing is specified.

**Screenshots**

**Version information**
Using latest version at time of writing, 72ba688
Installed via stack
Gtk version: 3.22.29+60+ge42d8598ca

Ping @u11gh

u11gh:
Fix #371: Search /etc/xdg even if XDG_CONFIG_HOME is set.  (As required by XDG specs.)

### DIFF
--- a/src/System/Taffybar/Information/XDG/DesktopEntry.hs
+++ b/src/System/Taffybar/Information/XDG/DesktopEntry.hs
@@ -24,6 +24,7 @@ module System.Taffybar.Information.XDG.DesktopEntry
   , deNotShowIn
   , deOnlyShowIn
   , existingDirs
+  , getDefaultConfigHome
   , getDefaultDataHome
   , getDirectoryEntriesDefault
   , getDirectoryEntry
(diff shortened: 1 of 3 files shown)


src/System/Taffybar/Information/XDG/DesktopEntry.hs
```diff
@@ -53,13 +53,21 @@ data DesktopEntryType = Application | Link | Directory
 existingDirs :: [FilePath] -> IO [FilePath]
 existingDirs  dirs = do
   exs <- mapM fileExist dirs
-  return $ S.toList $ S.fromList $ map fst $ filter snd $ zip dirs exs
+  let exDirs = nub $ map fst $ filter snd $ zip dirs exs
+  mapM_ (putStrLn . ("Directory does not exist: " ++)) $ dirs \\ exDirs
```
IvanMalison: Use printf... It's already imported.

PR #391, opened 2018-07-02
Removed Gtk2hs from FreedesktopNotifications

Tordek:


IvanMalison:
Looks like there is a build failure because you picked up a bad change that I added to master. If you merge the fixing commit it should pass.

Tordek:
Oops, the change for Windows went along for the ride... should I split it?

PR #393, opened 2018-07-02
Use gigtk by default

### LINKED ISSUE #256: Migrate from gtk2hs to gi-gtk
HELP WANTED! Choose a module, and change it to use gi-gtk. If you are looking for an example of how to do this, take a look at System.Taffybar.Widget.Windows
Modules that need to be converted:
- [ ] System.Taffybar.Widget
- [x] System.Taffybar.Widget.Battery
- [ ] System.Taffybar.Widget.CPUMonitor
- [x] System.Taffybar.Widget.CommandRunner
- [ ] System.Taffybar.Widget.Decorators
- [ ] System.Taffybar.Widget.DiskIOMonitor
- [ ] System.Taffybar.Widget.FSMonitor
- [ ] System.Taffybar.Widget.FreedesktopNotifications
- [ ] System.Taffybar.Widget.Generic.DynamicMenu
- [ ] System.Taffybar.Widget.Generic.Graph
- [ ] System.Taffybar.Widget.Generic.Icon
- [ ] System.Taffybar.Widget.Generic.PollingBar
- [ ] System.Taffybar.Widget.Generic.PollingGraph
- [x] System.Taffybar.Widget.Generic.PollingLabel
- [ ] System.Taffybar.Widget.Generic.VerticalBar
- [x] System.Taffybar.Widget.GitHubNotifications
- [ ] System.Taffybar.Widget.Layout
- [x] System.Taffybar.Widget.MPRIS2
- [ ] System.Taffybar.Widget.NetMonitor
- [ ] System.Taffybar.Widget.Generic.ChannelGraph
- [ ] System.Taffybar.Widget.NetworkGraph
- [x] System.Taffybar.Widget.SNITray
- [ ] System.Taffybar.Widget.SimpleClock
- [ ] System.Taffybar.Widget.Systray
- [ ] System.Taffybar.Widget.Text.CPUMonitor
- [ ] System.Taffybar.Widget.Text.MemoryMonitor
- [ ] System.Taffybar.Widget.Text.NetworkMonitor
- [ ] System.Taffybar.Widget.Util
- [ ] System.Taffybar.Widget.Volume
- [ ] System.Taffybar.Widget.Weather
- [x] System.Taffybar.Widget.Windows
- [x] System.Taffybar.Widget.Workspaces
- [ ] System.Taffybar.Widget.XDGMenu.Menu
- [ ] System.Taffybar.Widget.XDGMenu.MenuWidget
----
Original post below:
---- 
The [consensus](https://www.reddit.com/r/haskell/comments/5m9vlh/recommended_bindings_for_gtk_gtk2hs_vs_haskellgi/) of the haskell community seems to be that the generated gtk bindings offered by https://github.com/haskell-gi/haskell-gi should be preferred to the [gtk2hs](https://github.com/gtk2hs/gtk2hs)  bindings that taffybar currently uses.

I've been aware that we really ought to migrate for some time, but my attempts to solve #109 involve gi-gtk code, and could necessitate migrating taffybar to gi-gtk. Unfortunately, this migration will not be an easy one, since the interfaces to the two libraries are almost completely different. Further complicating the issue is the fact that the migration can't really be done incrementally AFAICT.

I want to start working on this migration immediately, but I want to get the thoughts of taffybar users and contributors about the best way to proceed.

Because incremental work is impossible given how deeply entrenched taffybar's dependency on gtk2hs is, I am considering creating an entirely new repository to house a greenfield implementation of the core functionality of taffybar using gi-gtk instead of gtk2hs. This is not an action I take lightly as it essentially amounts to a fork, which is an approach that I generally view as counterproductive, because it can be harmful to/divisive of the community. I sort suspect that at least initially, I would probably be the only one working on or using such a fork; even if I did eventually manage to attract some other users, it will probably be more of a slow trickle rather than a sudden exodus, and that situation (2 versions causing confusion and division) is obviously less than ideal.

One of the ideas that has been kicking around in my head is that it might be good to separate taffybar into core and contrib repositories/packages. This change would make migrating taffybar into a much more manageable task. The contrib migration would obviously take a little bit longer, but it could be done peicemeal. In this world there would be coexisting gi and gtk2hs versions of taffybar (ideally simply separated by a major version number) that would coexist until the migration is complete.

I guess I don't really have any specific questions, but I'd love to hear peoples thoughts. Is there a way that this could be approached that avoids some of the concerns I have? What do people think about taffybar core vs contrib? Would anyone be willing to help with the migration? Is anyone strongly opposed to migration?

Not really sure who to tag here, so feel free to add anyone:

@travitch @teleshoes @escherdragon @u11gh @sakshamsharma @iamjamestl @LeifW

Tordek:
Fixes #256


Tordek:
Welp, I'm not sure why `set` works in one version and not the other. That much should easily be fixed by just using the explicit `setWidgetSoAndSo`, I guess.

src/System/Taffybar/Widget/CommandRunner.hs
```diff
@@ -29,15 +30,15 @@ commandRunnerNew
   => Double -- ^ Polling period (in seconds).
   -> String -- ^ Command to execute. Should be in $PATH or an absolute path
   -> [String] -- ^ Command argument. May be @[]@
-  -> String -- ^ If command fails this will be displayed.
+  -> T.Text -- ^ If command fails this will be displayed.
   -> m GI.Gtk.Widget
 commandRunnerNew interval cmd args defaultOutput =
-  pollingLabelNew "" interval $
+  pollingLabelNew (T.pack "") interval $
```
IvanMalison: Use overloaded strings

src/System/Taffybar/Widget/FSMonitor.hs
```diff
@@ -30,12 +31,12 @@ fsMonitorNew
   -> [String] -- ^ Names of the partitions to monitor (e.g. [\"\/\", \"\/home\"])
   -> m GI.Gtk.Widget
 fsMonitorNew interval fsList = liftIO $ do
-  label <- pollingLabelNew "" interval $ showFSInfo fsList
+  label <- pollingLabelNew (T.pack "") interval $ showFSInfo fsList
```
IvanMalison: Use overloaded strings

src/System/Taffybar/Widget/Generic/VerticalBar.hs
```diff
@@ -103,8 +102,8 @@ liftedBarColor bc pct =
     BarConfig { barColor = c } -> return (c pct)
     BarConfigIO { barColorIO = c } -> c pct
 
-renderFrame :: Double -> BarConfig -> Int -> Int -> C.Render ()
-renderFrame pct cfg width height = do
+renderFrame_ :: Double -> BarConfig -> Int -> Int -> C.Render ()
```
IvanMalison: Why was the underscore added here?

IvanMalison:
Wow, thanks for taking all of this on -- it ended up being quite a bit of work. I took an (admittedly not super close) look at things and they look mostly fine, so I'm going to go ahead and merge. I'm sure some stuff will come up, so I'll hold off on doing a release until things seems stable. I think we need to tackle the segfault issue as well. Are you seeing it when you run taffybar?

src/System/Taffybar/Widget/Generic/VerticalBar.hs
Tordek: Because there's a Gtk.renderFrame already; I could have used `hiding` I guess.

Tordek:
I haven't seen any segfaults, but I haven't had it running for a long time either...

src/System/Taffybar/Widget/Generic/VerticalBar.hs
IvanMalison: Yes, that would probably be better. Thats part of why I prefer qualified imports.

IvanMalison:
@Tordek  @teleshoes I'm probably speaking too soon, but I haven't gotten any segfaults since I started running from this pull request. Haven't seen #228 in quite some time either

Tordek:
And here's a new one: SIGABRT `GLib:ERROR:/build/glib2.0-B1uXKV/glib2.0-2.50.3/./glib/gvarianttypeinfo.c:184:g_variant_type_info_check: assertion failed: (0 <= index && index < 24)`

Tordek:
From running it in `gdb`:

    0x00007ffff5b7649c in g_slice_alloc () from /lib/x86_64-linux-gnu/libglib-2.0.so.0

Tordek:
I keep getting errors in different places, so at the time I'm inclined to believe there's some use after free somewhere; probably on the polling Widgets.

Tordek:
Ok, I have one certainty at least: Spamming play/pause causes mpris2 to crash with a segfault. Changing only that in my config stops, at least, that source of crash.

IvanMalison:
@Tordek My guess has always been that it somehting to do with pxibufs/icons. which would be consistent with the mpris thing

IvanMalison:
@Tordek Continue discussion in #231

PR #401, opened 2018-07-09
the glib in version 2.0.17 was missing

fabianpage:
the glib in version 2.0.17 was missing when i tried to install with the quickstartscript


fabianpage:
actually that's not enough.

PR #415, opened 2018-09-08
Build with ghc-8.2.2 using Nix

PierreR:
This is not ready to be merged.

I would like to discuss this PR to get feedback and help.

This PR allows me to install `Taffybar` compiled with ghc-8.2.2. This is necessary because of https://github.com/taffybar/taffybar/issues/358.

You can try it out in nixos (and I guess also with nix and other linux distro) by using this command:
```
nix-env -i -f https://github.com/pierrer/taffybar/tarball/20584dafc952fbe5ee36136e0492500749ea7faa
```

It is going to take quite a while because a lot of the dependencies are not in the nixos default binary cache.

One issue remains. It works on the stable nixos channel (18.03) but it won't work so well when I use the unstable channel (18.09). In `18.09` it completes the building process without problem but Taffybar will fail to recompile the user configuration. It complains that it can't find the imported taffybar modules.

As a bonus, I would also like to speed up the build/compilation process (by disabling profiling globally, using `justStaticExecutables´, ...). Any idea or suggestion welcomed.

When I am happy about the workflow I will probably push it to [cachix](https://cachix.org/)

Thanks for your help and feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .stack-work
 .cabal-sandbox
 cabal.sandbox.config
+/result
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,38 @@
+# You can build this repository by running:
+#   $ nix-build
+
+{
+pkgs_path ? ./nix/nixpkgs.nix
+}:
+
+let
+  overlay = import ./nix/overlay.nix;
+  pkgs = import pkgs_path {
+    config = {};
+    overlays = [ overlay ];
+  };
+  filter =  import ./nix/filter.nix {inherit (pkgs)lib;};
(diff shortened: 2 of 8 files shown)


IvanMalison:
Thanks for submitting, this should be helpful to a lot of people. I'll try it out on my machine in the near future.

IvanMalison:
@PierreR I'm confused. Why does this need to exist at all. Seems like we should just use the existing nixpkgs derivations for taffybar.

PierreR:
Yep it can be removed now that everything is working correctly in nixpkgs

PierreR:
I can do a PR to clean this up.

PR #417, opened 2018-09-20
markupEscapeText expects length in bytes, not characters

### LINKED ISSUE #416: Window titles containing multibyte characters are chopped off
When a window title includes multibyte characters, the string is displayed prematurely truncated.  For example, the Unicode 'dot' in this title results in the 'x' in Firefox being chopped off:

taffybar-truncated-window-title

This seems to be because when the strings are passed to the [`markupEscapeText`](https://hackage.haskell.org/package/gi-glib-2.0.17/docs/GI-GLib-Functions.html#g:171) function, the length of the text in characters is being passed when the function really wants the length of the text in bytes.  Strings with multibyte characters will have more bytes than characters, thus the size is understated and the ends of the strings get lopped off.

iamjamestl:
When text contains multi-byte characters, the length is undercalculated
for the `markupEscapeText` function, which expects the length of the
text in bytes.  This results in some labels being prematurely truncated.

I began replacing calls to `Data.Text.length` with
`Data.ByteString.length` when I read that `markupEscapeText` can also
accept the length as `-1` when the text is null-terminated, which I have
found to be the case in all the places it's used.  This approach doesn't
add a new dependency on the `bytestring` package and may be less
expensive.

Closes #416

### DIFF
--- a/src/System/Taffybar/Widget/FreedesktopNotifications.hs
+++ b/src/System/Taffybar/Widget/FreedesktopNotifications.hs
@@ -121,8 +121,8 @@ notify s appName replaceId _ summary body _ _ timeout = do
                            Nothing -> Just timeout
                            Just maxTimeout -> Just (min maxTimeout timeout)
 
-  escapedSummary <- markupEscapeText summary (fromIntegral $ T.length summary)
-  escapedBody <- markupEscapeText body (fromIntegral $ T.length body)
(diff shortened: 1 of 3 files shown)


IvanMalison:
Hmm this has conflicts. Might just be with the other change you made

iamjamestl:
Resolved the merge conflict.

Issue #379, opened 2018-06-13
Windows widget: HTML de-encoded in menu label but not active label


ARR8:
**Describe the bug**
Not sure if this is a known problem. Doesn't seem to be in the docs about the window widget.  
A window title with HTML-encoded characters shows up normally in the active label, but changes to HTML in the window list, in some cases expanding the width dramatically.

**To Reproduce**
Open a program with such characters in the title.

**Expected behavior**
Menu label same as active label.

**Screenshots**
![screenshot2](https://user-images.githubusercontent.com/8064135/41384369-dbe26afa-6f64-11e8-8001-385289047e1b.png)

**Version information**
Latest version, afbe492
Installation method (nix/stack/cabal/arch packages): Stack
Gtk version: 3.22.29+60+ge42d8598ca

IvanMalison:
Hmmm not sure what this could be. Both default markup functions escape markup in the same way so its not clear to me what would be causing this.

ARR8:
Maybe something with GTK itself?

IvanMalison:
@ARR8 Is this still an issue @Tordek changed the way things are escaped.

ARR8:
Yes, this is still an issue

IvanMalison:
@iamjamestl This is fixed now, correct?

@ARR8 Can you verify?

iamjamestl:
@IvanMalison, the window widget has been working for me as expected since #418 was merged.

PR #418, opened 2018-09-20
windows: Fix over-escaped text in menu labels

iamjamestl:
It was observed in #379 that the window widget menu labels appear to be
wrongly escaped while the active label is OK.  This appears to be
because the GTK menuitem labels don't support markup, or at least
there seems to be no way to access the internal label to enable the
markup property.  Because of this limitation, there is no need to escape
the menuitem label text, as implemented in this change.


IvanMalison:
@iamjamestl thanks for gettring to the bottom of this!

PR #424, opened 2018-10-29
Bugfix: Expire time in millisecond is wrongly taken as second.

Magicloud:
Per https://developer.gnome.org/notification-spec/, expire timeout value
passed in is millisecond. Since threadDelay takes microsecond, the
multiply times should be 10^3, instead of 10^6.


Magicloud:
Sorry, not sure if you have merged or fixed this.

Magicloud:
My bad. Closing.

Issue #382, opened 2018-06-18
How to style battery icon?


purefn:
I've updated to taffybar 2.1.1 and am working on getting my setting moved over. So far things are going mostly ok. One thing I haven't been able to figure out is how to style certain elements. In this case, the `batteryIconNew`. I changed the background of my bar to black, which makes the battery no longer visible. I'd like to change the battery icon to be white, but haven't figured out how to do that yet. Any tips on how to figure out what CSS classes to use for the various icons and things?

IvanMalison:
@purefn The current batteryIcon widget simply uses the icon name provided by the upower dbus interface to load the appropriate pixbuf using gtk's icon loading machinery.

Because it is just loading a pixbuf (basically an image), I'm not sure that much can be done with css to change the appearance of the icon. You can swap out the icon set by making changes to the global icon theme though.

Obviously, this is not ideal. I would like to add a battery icon widget that is drawn dynamically with cairo (taffybar used to have something like this, but the code was a mess and it didn't really look good).

IvanMalison:
@purefn If you want something that is like what used to exist you can use

channelWidget and verticalBar to whip something up pretty easily. Doesn't look very good though, IMO.

purefn:
Thanks for the quick response. I will probably look into using something that can go in the tray. [battery-monitor](https://github.com/maateen/battery-monitor) looks promising, I'll just have to get it into nixpkgs.

PR #425, opened 2018-11-25
Allow batteryIcon styling (issue #382)

amurrayc:
This allows the battery icon to take its colour from `taffybar.css`. By
default it uses the `color` attribute from the top `.taffy-window` class
but can be overridden in the `.auto-size-image` class.

Related to #382

### DIFF
--- a/src/System/Taffybar/Widget/Battery.hs
+++ b/src/System/Taffybar/Widget/Battery.hs
@@ -109,12 +109,16 @@ batteryIconNew = do
   ctx <- ask
   liftIO $ do
     image <- imageNew
+    styleCtx <- widgetGetStyleContext =<< toWidget image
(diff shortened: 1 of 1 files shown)


IvanMalison:
@amurrayc Very cool. Nice work!

IvanMalison:
Not sure I would say this fixes #382 since this still seems to use static images.

amurrayc:
Agreed. However, that issue started as a bugreport (can't get battery icon to be white) and then expanded into a feature request (dynamically drawn battery widget).

This fixes the bug but does nothing for the feature request. :)

IvanMalison:
> Agreed. However, that issue started as a bugreport (can't get battery icon to be white) and then expanded into a feature request (dynamically drawn battery widget).

Ahh didnt realize that

Issue #352, opened 2018-05-19
Use Proper Broadcast channel/0 reader spaceleak


IvanMalison:
such as https://github.com/merijn/broadcast-chan

the current approach (that uses dupChan to simulate a broadcast channel has a zero reader space leak)

amurrayc:
What would be the desired behaviour for shared channels?  At the moment, each widget (at least the ones created by `channelWidgetNew` and `channelGraphNew`) has its own channel that connects to an information source.  Would the idea be to allow two related widgets (like a battery icon and a battery text widget) to share an information source?

Wouldn't most widgets have independent sources?

IvanMalison:
> What would be the desired behaviour for shared channels? At the moment, each widget (at least the ones created by `channelWidgetNew` and `channelGraphNew`) has its own channel that connects to an information source. Would the idea be to allow two related widgets (like a battery icon and a battery text widget) to share an information source?

As I mentioned in your pull request, this is not correct. They all share a channel because the value is retrieved through `getStateDefault`, which ensures that there is only one value that is retrieved every time whatever the relevant get function is called.

IvanMalison:
> Wouldn't most widgets have independent sources?

Yes, most widgets have independent sources (although not all e.g. battery graph or icon and battery percentage), but people with multiple displays (and multiple taffybars) are going to have multiple instances of the same widget. The current implementation uses a channel and a single source for information retrieval, but updates all of the widgets for each new piece of data.

IvanMalison:
> They all share a channel because the value is retrieved through `getStateDefault`, which ensures that there is only one value that is retrieved every time whatever the relevant get function is called.

I should clarify that they each have their own read end of the channel, but they share the write end of the channel.

IvanMalison:
@amurrayc if you want to spearhead migrating all of our channels to https://github.com/merijn/broadcast-chan I would greatly appreciate it.

I think the reason I didn't take it on myself was that I was a little worried that it doesn't seem to have all that much community support, but maybe we can help with that I guess.

amurrayc:
I've done a pretty simple switch to broadcast-chan in the abovementioned fork. It's in the broadcast-chan branch.

As far as I can tell, in the Chan-based approach, the first listener creates the information channel but then it, and all subsequent listeners `dupChan` and then start reading. That means there are n+1 channels but only n listeners. Am I missing something?

[...]

PR #426, opened 2018-12-04
Remove unnecessary, leaky Chan duplication

amurrayc:
Related to #352:
The `dupChan` in `channelWidgetNew` and `channelGraphNew` are
unnecessary since all widgets so created (battery and network) create
new channels.  These are then duplicated and the original is left
hanging, accumulating unread writes.  This eliminates a big space leak,
especially for the battery widgets.

### DIFF
--- a/src/System/Taffybar/Widget/Generic/ChannelGraph.hs
+++ b/src/System/Taffybar/Widget/Generic/ChannelGraph.hs
@@ -12,9 +12,8 @@ channelGraphNew
 channelGraphNew config chan sampleBuilder = do
   (graphWidget, graphHandle) <- graphNew config
   _ <- onWidgetRealize graphWidget $ do
-       ourChan <- dupChan chan
        sampleThread <- forkIO $ forever $ do
-         value <- readChan ourChan
+         value <- readChan chan
          sampleBuilder value >>= graphAddSample graphHandle
        void $ onWidgetUnrealize graphWidget $ killThread sampleThread
   return graphWidget
--- a/src/System/Taffybar/Widget/Generic/ChannelWidget.hs
+++ b/src/System/Taffybar/Widget/Generic/ChannelWidget.hs
@@ -8,9 +8,8 @@ import GI.Gtk
 channelWidgetNew :: (MonadIO m, IsWidget w) => w -> Chan a -> (a -> IO ()) -> m w
 channelWidgetNew widget channel updateWidget = do
   void $ onWidgetRealize widget $ do
-    ourChan <- dupChan channel
     processingThreadId <- forkIO $ forever $
-      readChan ourChan >>= updateWidget
+      readChan channel >>= updateWidget
     void $ onWidgetUnrealize widget $ killThread processingThreadId
   widgetShowAll widget
   return widget


amurrayc:
I agree that this does not directly address that issue but it is relevant, I think.

I am looking into a broadcast-chan based approach but as I investigated the code I noticed this anomaly.  I am getting a large space leak with the batteryIcon widget and a lesser one with the networkGraph one so I decided it would be good and (I think) harmless to push this change for now.

For `batteryIconNew`: (the other three are similar)

`batteryIconNew` calls `getDisplayBatteryChan` which ends up calling `monitorDisplayBattery` which calls `newChan`.  This new `Chan` is then used by `batteryIconNew` in a call to `channelWidgetNew` which calls `dupChan`.

Grepping for `channelWidgetNew` and `channelGraphNew` show that they are only used in `batteryIconNew`, `textBatteryNew`, `networkMonitorNew` and `networkGraphNew`.

I suppose there could be a problem with user-defined widgets that don't conform to this pattern but, at least for these widgets (especially the battery ones) the leak is prohibitive.  This change lets me use taffybar while I look into broadcast-chan.

I'll comment more on that under the issue.

IvanMalison:
I suspect that this pull request will not work correctly when you display taffybar on multiple monitors with the same widgets on each monitor.

IvanMalison:
> I am getting a large space leak with the batteryIcon widget and a lesser one with the networkGraph one so I decided it would be good and (I think) harmless to push this change for now.

I really don't think so. Maybe I am misunderstanding something, but I think that it might be that you have misunderstood how the channel mechanism is supposed to work. Let me know if my comments have clarified things at all for you.

Btw. I don't get a space leak atm, how are you sure about the cause of the space leak?

amurrayc:
Okay.  I see.

There still remains the issue that the first (and quite possibly only) widget which creates the channel sets up the information source using this channel before calling `dupChan` and attaching the reader.  As far as I can tell, this still leaves a dangling channel with no reader.

amurrayc:
Okay.  I understand now the issue.  Sorry for the noise.  Absolutely, your
comments make it clear.  I've just started digging in and found this
problem before I had a chance to grok properly.

I get the space leak when I use either of the battery widgets (so it's not
my other PR :). It's fairly dramatic.  As soon as I remove the duplicated
channel it disappears.  I'm using a fairly lightly modified version of the
taffybar.hs.example file.

I'm pretty certain about the leak.  Profiling shows a buildup in
getBatteryInfo and unmarshalMessageM on the system bus.

I'll try a standard stack build tonight to see if there is any difference.

I am compiling taffybar with ghc 8.4.4 using nix but I also had a similar
(I suspect the same) leak with 8.2.2.

IvanMalison:
> As far as I can tell, this still leaves a dangling channel with no reader.

Right, this is exactly the issue described by #352, but (my understanding is that) its fine, so long as at least one channel is actually reading from the base channel. Perhaps this is not the case.

LeifW:
I was wondering if it actually needs to be a channel, or could it just be a
single cell that the widgets read the current value from?

IvanMalison:
> I was wondering if it actually needs to be a channel, or could it just be a single cell that the widgets read the current value from?

The channel works better for things like a graph where you want to display some unspecified number of historical values as well. Without a channel, it would be too easy to drop values if a read did not happen at the right time.

amurrayc:
Okay, so this PR is not right.  I've a better solution in my fork at https://github.com/amurrayc/taffybar in the broadcast-chan branch.

New to github flow: what should I do? Close this PR?

IvanMalison:
@amurrayc MAke a new pull request. Reference that one here if you want.

PR #430, opened 2018-12-09
Allow CSS stylesheet to be specified from TaffybarConfig

bgamari:
This makes it easier to package a taffybar configuration into a
self-contained Cabal package.

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -154,8 +154,8 @@ getDefaultConfigFile name = do
   dataDir <- getDataDir
   return (dataDir </> name)
 
-startCSS :: IO Gtk.CssProvider
-startCSS = do
+startCSS :: Maybe FilePath -> IO Gtk.CssProvider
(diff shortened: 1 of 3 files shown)


IvanMalison:
@bgamari Cool thanks.

As it stands, this will still load the css from the default location (the taffybar directory), and that style sheet will take precedence because it is loaded afterwards. Was this your intention? Why does this seem like the right thing to do?

I'm still thinking about it, but I might change the precedence order, or not load the default css location at all. What do you think?

bgamari:
Ivan Malison <notifications@github.com> writes:

> @bgamari Cool thanks.
>
> As it stands, this will still load the css from the default location (the taffybar directory), and that style sheet will take precedence because it is loaded afterwards. Was this your intention? Why does this seem like the right thing to do?
>
> I'm still thinking about it, but I might change the precedence order, or not load the default css location at all. What do you think?
>
For what it's worth either ordering is fine with me.

PR #440, opened 2019-02-19
add CSS class to `Layout`

dlahoti:
Gives the label in `Layout` the class `layout-label` so that it can be styled with CSS.

### DIFF
--- a/src/System/Taffybar/Widget/Layout.hs
+++ b/src/System/Taffybar/Widget/Layout.hs
@@ -31,6 +31,7 @@ import           GI.Gdk
 import           System.Taffybar.Context
 import           System.Taffybar.Information.X11DesktopInfo
 import           System.Taffybar.Util
+import           System.Taffybar.Widget.Util
 
 -- $usage
 --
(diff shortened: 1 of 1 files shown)


IvanMalison:
Was about to merge, but noticed the build failed. I think it might be a really weird flake, but can't be sure. any ideas?

IvanMalison:
Wait, nvm, looks like the error is here:

https://travis-ci.org/taffybar/taffybar/builds/495689768#L1203

Seems like you need a text value rather than a string value.

How are you building this locally? You might be using a different version of gdk or gtk bindings?

dlahoti:
Looks like I forgot to add `OverloadedStrings` – my local version has some other changes, so I had just manually copied over the relevant ones here.

IvanMalison:
@dlahoti What changes? Are they not worth merging?

dlahoti:
I have the local version in my `.config/taffybar/lib/` folder, so it has various name changes to avoid conflicting with the existing version.

PR #449, opened 2019-04-28
Add new widget SimpleCommandButton

### LINKED ISSUE #420: Simple Command Button Widget
I want to add a button to my taffybar that simply executes a command (in my case, to rotate the screen). I am surprised that there is no readily available widget for that – `System.Taffybar.Widget.CommandRunner` sounded promising, but is something different.

Expecting a type signature like
```
simpleButton :: MonadIO m
  => String --  Command to execute. Should be in $PATH or an absolute path
  -> String -- Label for the button
  -> m Widget
```

u11gh:
Fixes #420

### DIFF
--- a/src/System/Taffybar/Widget.hs
+++ b/src/System/Taffybar/Widget.hs
@@ -44,6 +44,9 @@ module System.Taffybar.Widget
   , defaultClockConfig
   , ClockConfig(..)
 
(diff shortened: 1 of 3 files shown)


IvanMalison:
@u11gh Do you mind adding entries to the changelog for this and the other pull request?

PR #472, opened 2019-06-30
Bugfix/leaking update samples

RubenAstudillo:
When using `taffybar`, I saw it leaking memory under heavy network loads. The graph produced was the following 
taffybar-linux-x86_64-before. It showed the problem was around the functions `parseDevNet` and `updateSamples`. This was also confirmed by the retainer graph profiling. 
taffybar-linux-x86_64-before3
 Adding `seq` on certain places i could reduce the memory consumption by those closures as shown on the following retainer profile graph.
taffybar-linux-x86_64
and on the cost center profile.
taffybar-linux-x86_64

### DIFF
--- a/src/System/Taffybar/Information/Network.hs
+++ b/src/System/Taffybar/Information/Network.hs
@@ -55,7 +55,7 @@ getDeviceUpDown s = do
   dev <- initSafe <$> s `atMay` 0
   down <- readDef (-1) <$> s `atMay` 1
   up <- readDef (-1) <$> s `atMay` out
-  return (dev, (down, up))
+  dev `seq` down `seq` up `seq` return (dev, (down, up))
   where
     out = length s - 8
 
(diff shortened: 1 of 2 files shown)


IvanMalison:
@RubenAstudillo Do you mind adding an entry to the changelog for this?

src/System/Taffybar/Information/Network.hs
```diff
@@ -103,7 +103,7 @@ monitorNetworkInterfaces interval onUpdate = void $ do
         let speedInfo = map sampleToSpeeds samples
         onUpdate speedInfo
         return samples
-      doUpdate = MV.modifyMVar_ samplesVar ((>>= doOnUpdate) . updateSamples)
+      doUpdate = MV.modifyMVar_ samplesVar (doOnUpdate <=< updateSamples)
```
IvanMalison: Is this change necessary? How does this impact anything?

RubenAstudillo: Not really, I thought it was clearer, but I will roll it back!.

RubenAstudillo:
> @RubenAstudillo Do you mind adding an entry to the changelog for this?

OK, should this be under a 3.2.2 header or still on the 3.2.1 header?

src/System/Taffybar/Information/Network.hs
RubenAstudillo: done on rebase!

PR #492, opened 2020-07-09
Add SNITray-related documentation and fix formatting

quarkQuark:
Add haddock documentation to System.Taffybar and System.TaffyBar.Widget.SNITray to clarify some things I struggled with recently in setting up taffybar. Namely, that only the SNI protocol is supported, and that status-notifier-watcher needs to be run.

Additionally, I fixed a couple of errors in the formatting of System.Taffybar that were stopping the Haddock documentation from displaying correctly.

### DIFF
--- a/src/System/Taffybar.hs
+++ b/src/System/Taffybar.hs
@@ -25,7 +25,7 @@ module System.Taffybar
   -- like reserving strut space so that window managers don't put windows over
   -- it).
   --
-  -- | The config file in which you specify the gtk+ widgets to render is just a
+  -- The config file in which you specify the gtk+ widgets to render is just a
   -- Haskell source file which is used to produce a custom executable with the
   -- desired set of widgets. This approach requires that taffybar be installed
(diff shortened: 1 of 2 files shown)


IvanMalison:
@quarkQuark Looks like there is a haddock issue: https://travis-ci.org/github/taffybar/taffybar/builds/706689983

quarkQuark:
I managed to get rid of the haddock error, but there's one more error left in the automated check that I don't understand. Something to do with nix and permissions?

```
$ cachix use taffybar
This user doesn't have permissions to configure binary caches.
You can either:
a) Run the same command as root to configure them globally.
b) Run the following command to add your user as trusted 
   and then try again:
  echo "trusted-users = root travis" | sudo tee -a /etc/nix/nix.conf && sudo pkill nix-daemon
The command "cachix use taffybar" exited with 1.
```

MyriaCore:
Bump? This still appears relevant

PR #495, opened 2020-08-24
Add named icons to System.Taffybar.Widget.Generic.Icon

MyriaCore:
This PR adds named icons to the generic icon widget that ships with taffybar. This would allow for much easier creation of widgets involving icons, as users could just use icons from their GTK theme, and avoid having to homebrew their own versions of this widget to do so. 

Adds 2 new functions - `iconImageWidgetNewFromName` and `pollingIconImageWidgetNewFromName`. 

As an example, `iconImageWidgetNewFromName "weather-clouds"` yields an icon that looks like image. `pollingIconImageWidgetNewFromName` works similarly to `pollingIconImageWidgetNew`, but as with `iconImageWidgetNewFromName`, you get the added benefit of specifying via icon names as opposed to paths. 

Most of the source here was taken from [my personal modifications](https://gitlab.com/myriacore-dotfiles/taffybar/-/blob/ad1c1919ac840a295b190179f7e334e544b22f37/Widget/Generic/Icon.hs) of `System.Taffybar.Widget.Generic.Icon`.

### DIFF
--- a/src/System/Taffybar/Widget/Generic/Icon.hs
+++ b/src/System/Taffybar/Widget/Generic/Icon.hs
@@ -29,7 +29,8 @@ iconImageWidgetNew path = liftIO $ imageNewFromFile path >>= putInBox
 -- names are sourced from the current GTK theme. 
 iconImageWidgetNewFromName :: MonadIO m => T.Text -> m Widget
 iconImageWidgetNewFromName name = liftIO $ 
-  imageNewFromIconName (Just name) (fromIntegral $ fromEnum IconSizeMenu) >>= putInBox
+  imageNewFromIconName (Just name) (fromIntegral $ fromEnum IconSizeMenu) 
+  >>= putInBox
 
 -- | Create a new widget that updates itself at regular intervals.  The
 -- function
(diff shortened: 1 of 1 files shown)


MyriaCore:
`stack build` seems to work on my side. Not sure why I'm failing the cachix tests. In both tests, it seems to be a user permissions thing:

> ```
> $ cachix use taffybar
> 
> This user doesn't have permissions to configure binary caches.
> 
> You can either:
> 
> a) Run the same command as root to configure them globally.
> 
> b) Run the following command to add your user as trusted 
> 
>    and then try again:
> 
>   echo "trusted-users = root travis" | sudo tee -a /etc/nix/nix.conf && sudo pkill nix-daemon
> 
> The command "cachix use taffybar" exited with 1.
> ```

As I said, the package builds for me. I've been runing this code for a while, but I haven't really tested it.

src/System/Taffybar/Widget/Generic/Icon.hs
```diff
@@ -49,6 +63,37 @@ pollingIconImageWidgetNew path interval cmd = liftIO $ do
...
+    _ <- forkIO $ forever $ do
+      let tryUpdate = do
+            name' <- cmd
+            postGUIASync $
+              imageSetFromIconName icon (Just name') $
+                fromIntegral $ fromEnum IconSizeMenu
+      E.catch tryUpdate ignoreIOException
+      threadDelay $ floor (interval * 1000000)
+    return ()
+  putInBox icon
```
IvanMalison, line 95: This repeats too much code from the above function. The only lines that are different here are 90 and 84. It should be pretty easy to factor those part out so they are passed in as a parameter to a more general function that can handle both cases. You can keep the new functions names as simple aliases that just call the more general function appropriately.

MyriaCore: Will do!

MyriaCore: Resolved via ec2f227

MyriaCore:
@IvanMalison hey, I did the refactor!

MyriaCore:
bump? this was kinda a while ago.

PR #514, opened 2021-07-31
Add tooltip to network graph.

kurnevsky:
Tooltip with `defaultNetFormat` from `NetworkMonitor`.

### DIFF
--- a/src/System/Taffybar/Widget/NetworkGraph.hs
+++ b/src/System/Taffybar/Widget/NetworkGraph.hs
@@ -1,11 +1,16 @@
 module System.Taffybar.Widget.NetworkGraph where
 
+import Control.Monad
 import qualified GI.Gtk
+import GI.Gtk.Objects.Widget (widgetSetTooltipMarkup)
 import System.Taffybar.Context
 import System.Taffybar.Hooks
 import System.Taffybar.Information.Network
+import System.Taffybar.Util (postGUIASync)
 import System.Taffybar.Widget.Generic.ChannelGraph
+import System.Taffybar.Widget.Generic.ChannelWidget
 import System.Taffybar.Widget.Generic.Graph
+import System.Taffybar.Widget.Text.NetworkMonitor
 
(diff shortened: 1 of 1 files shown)


src/System/Taffybar/Widget/NetworkGraph.hs
```diff
@@ -20,4 +25,9 @@ networkGraphNew config interfaces = do
       toLogScale = logScale 2 (2 ** 32)
       toSample (up, down) = map (toLogScale . fromRational) [up, down]
       sampleBuilder = return . toSample . getUpDown
-  channelGraphNew config chan sampleBuilder
+  widget <- channelGraphNew config chan sampleBuilder
+  void $ channelWidgetNew widget chan $ \speedInfo ->
+    let (up, down) = sumSpeeds $ map snd speedInfo
+        tooltip = showInfo defaultNetFormat 3 (fromRational down, fromRational up)
```
IvanMalison, line 31: shouldn't we let this format be configurable?

kurnevsky: Yes, but for this we need to change the signature of `networkGraphNew`. Is it ok? Or probably create a new function.

IvanMalison: okay this confused me for a very long time when reading this:

When you name something like this tooltip, it makes it seem like it is some type of widget. The name should make it clear that this is just a string

IvanMalison (approved):

src/System/Taffybar/Widget/NetworkGraph.hs
IvanMalison: > Yes, but for this we need to change the signature of `networkGraphNew`. Is it ok? Or probably create a new function.

Right just add a parameter and rename this function. Then make a new function with the same name as this one and provide the default configuration calling the function with the added parameter.

kurnevsky: Added new function with config.

PR #523, opened 2021-08-13
Make MPRIS widget more tolerant of missing xesam:artist

NickHu:
This makes `getSongInfo` not fail when there is missing `xesam:artist` metadata exposed. In particular, I noticed that when using https://github.com/hoyon/mpv-mpris to play YouTube videos, this metadata is missing. This means that the widget does not draw. I also changed the view function to prevent it from displaying ` - %{title}` in the absence of artists.


IvanMalison:
@NickHu would you mind adding a changelog entry?

NickHu:
@IvanMalison Done in #524

PR #529, opened 2021-08-20
Catch exceptions from icon getters.

### LINKED ISSUE #403: Taffybar crashes when window requires icon that is not found in icon theme
Errors will look like:

```
taffybar: Icon 'audio-volume-medium' not present in theme breeze (0)
```

Setting 

```
getWindowIconPixbuf = scaledWindowIconPixbufGetter getWindowIconPixbufFromEWMH
```

in WorkspacesConfig can circumvent this issue for now, but we should find a better long term fix.

kurnevsky:
Fixes #403

### DIFF
--- a/src/System/Taffybar/Widget/Workspaces.hs
+++ b/src/System/Taffybar/Widget/Workspaces.hs
@@ -17,6 +17,7 @@ import           Control.Applicative
 import           Control.Arrow ((&&&))
 import           Control.Concurrent
 import qualified Control.Concurrent.MVar as MV
+import           Control.Exception.Enclosed (catchAny)
 import           Control.Monad
 import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Class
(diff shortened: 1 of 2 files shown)


src/System/Taffybar/Widget/Workspaces.hs
```diff
@@ -700,9 +705,9 @@ defaultGetWindowIconPixbuf =
 
 unscaledDefaultGetWindowIconPixbuf :: WindowIconPixbufGetter
 unscaledDefaultGetWindowIconPixbuf =
-  getWindowIconPixbufFromDesktopEntry <|||>
-  getWindowIconPixbufFromClass <|||>
-  getWindowIconPixbufFromEWMH
+  handleIconGetterException getWindowIconPixbufFromDesktopEntry <|||>
```
IvanMalison, line 708: @kurnevsky just thinking out loud here, but might it be possible to apply this automatically as part of <|||> or something?

IvanMalison: If not, I think I would prefer it if these were part of the definition of `getWindowIconPixbufFromDesktopEntry` etc. instead of it only being applied to the default.

IvanMalison: also, especially in the case of `getWindowIconPixbufFromDesktopEntry` I think we should log a warning that even though there was an icon to apply for the window, we could not apply it because the icon was missing from the theme.

Without something like that, it would likely be difficult for people to work out that their icon theme is not working for some reason.

kurnevsky: > as part of <|||>

In this case it won't work with a single getter. I can apply it to the definition of these functions though.

IvanMalison: > In this case it won't work with a single getter. I can apply it to the definition of these functions though.

Yeah i think maybe this is the best course of action. I think with scaledPixbufGetter and now this, there are just too many random wrapping functions to add to these pixbuf getters, and that stuff is likely going to confuse most users.

IvanMalison: > In this case it won't work with a single getter. I can apply it to the definition of these functions though.

well im not saying erase the existing function, just apply it automatically as part of <|||>

kurnevsky: But then it won't work if user defines single getter with no usages of `<|||>`.

kurnevsky: Instead I just applied it to each of these getters.

Issue #345, opened 2018-05-18
Graph colors can't be themed with CSS


colonelpanic8:
This is because they are drawn directly with cairo.

It would be nice to make it possible

IvanMalison:
https://hackage.haskell.org/package/gi-gtk-3.0.23/docs/GI-Gtk-Functions.html#v:renderBackground is also a possibility

IvanMalison:
`styleContextGetColor` could be used for the foreground color

PR #531, opened 2021-09-05
Use Data.Default for Default Values

### LINKED ISSUE #518: Use Data.Default for default values everywhere in taffybar
... not much more to say

pnotequalnp:
Fixes #518

I'm not sure which of the last five commits should be merged, but I put them here so it can be decided. The first two of them update the examples to use `def` where possible, the third updates the executable to use it, and the last two change argument names which were previously `def` and made GHC emit a warning for shadowing with the importing of `Data.Default.def`.

There are also several defaults exported which cannot be replaced with `def` because they are not simple data types but rather functions or type aliases. Notably, `defaultMPRIS2Config :: MPRIS2Config MPRIS2PlayerWidget` can technically be done but would require `FlexibleInstances` and for the user to supply additional type information at the use site (with an annotation or type application), so I left it out.

If it's all good I can squash it down to one commit for merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@
    tooltips.
 
  * Swap variables are added to `MemoryMonitor`.
+ 
+ * Many types have `Default` instances.
 
 ## Changes
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -3,6 +3,7 @@
 -- recompile itself.
 module Main ( main ) where
 
+import Data.Default (def)
 import Data.Semigroup ((<>))
 import Data.Version
 import Options.Applicative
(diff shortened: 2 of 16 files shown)


IvanMalison:
@pnotequalnp can you add a changelog entry?

IvanMalison:
Appreciate the PR. Curious why you decided to make this change.

pnotequalnp:
> Curious why you decided to make this change.

Just picked an easy issue to do to get more familiar with the codebase.

IvanMalison:
> Just picked an easy issue to do to get more familiar with the codebase.

Ah cool. An issue that might be easy AND useful is #345 .

Issue #399, opened 2018-07-05
Workspaces: Highlight the workspace being displayed on the workspace that the taffybar window is displayed on


IvanMalison:
This means that different taffybar windows will highlight different workspaces if they are displayed on different monitors.

PR #536, opened 2021-10-20
Implement focused-monitor CSS class

echarliewhite:
This pull request adds css classes which indicate whether or not a taffybar
window is on the currently focused monitor (see #399).

This is working on my setup (XMonad, two monitors, one taffybar on each), but
has not been tested otherwise. There are a few issues that should probably be
addressed before merging (see below).

Implementation notes (short version)
- Each taffybar window subscribes to `_NET_CURRENT_DESKTOP` changes
- on a desktop change, taffybar window location is compared against the active
  monitor's geometry, which is acquired from XMonad via pagerHints
- The top-level window is given a css class of either "focused-monitor" or
  "unfocused-monitor"

Implementation notes (long version)
- The general approach is copied from Workspaces.hs
- Currently each taffybar window subscribes to `_NET_CURRENT_DESKTOP` changes
  upon creation (in `buildBarWindow`), and will unsubscribe if the window is
  destroyed.
  - These separate subscriptions could probably be merged into one subscription
    in `buildContext`, but I was unsure about how to unsubscribe if a context
    is destroyed (while the taffybar process continues to run). I'm guessing
    this never happens, but wasn't sure.
- A new hint `_XMONAD_FOCUSED_MONITOR_GEOMETRY` is added to PagerHints
  - it contains XMonad's focused screen geometry as a string with format
    "x y w h"
  - I had to use screen geometry because XMonad's `ScreenID` and taffybar's
    `monitorNumber` didn't always match up (not sure why)
  - non-XMonad implementations might use `_NET_DESKTOP_LAYOUT_Sn` or
    `_NET_WORK_AREA`
- small stuff:
  - `rateLimitFn` and `updateWidgetClasses` are copied from Workspaces.hs,
    could move these to Util.hs, since they are shared
  - the css classes and rateLimit are currently hardcoded
  - I'm new at haskell so most of this is a combination of trial-and-error and
    copy-pasting. `updateTaffyWindowStatus` in particular should be closely
    reviewed 😬


IvanMalison:
I don't think any additional code is needed for this. There is already a css class that should get applied to the active window.

echarliewhite:
Or I could just put it in Workspaces, if the "no workspace widget" use case isn't a concern

IvanMalison:
> I thought about putting it in Workspaces, but it occurred to me that someone might want to make use of the css class on a bar that doesn't have a workspace widget.

Yeah, that makes sense.



> small stuff:
> * `rateLimitFn` and `updateWidgetClasses` are copied from Workspaces.hs,
>   could move these to Util.hs, since they are shared

Yeah, this is a huge redflag.

The right way to do this is to decouple the representation of stuff related ewmh from the workspace widget, and then provide updates to our internal view of ewmh state over a channel.

This would allow all workspaces to share the same subscription, and also enable your use case of having widgets other than workspaces have access to ewmh state.

echarliewhite:
Thanks for taking a look at this :)

echarliewhite:
> I seriously doubt that we need rateLimit here. Thats necessary in workspaces, but in this case, we should just be able to unconditionally listen on all events. Doubt we will get spammed.
Don't copy updateWidgetClasses...

👍

> I believe that this is because ScreenId is either internal to XMonad or an X11 property, while monitorNumber should be something that is obtained from gdk. I still think there should a way to establish a mapping between them that makes sense. Did you look in to this at all? In general pagerHints is a hack that we should avoid extending or overusing.

As far as I can tell, gdk is using X11's numbering system (https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/x11/gdkscreen-x11.c#L1067). XMonad applies its own numbering (https://hackage.haskell.org/package/xmonad-0.15/docs/src/XMonad.StackSet.html#new), but I think it matches Xinerama's numbering (https://gitlab.freedesktop.org/xorg/lib/libxinerama/-/blob/master/src/Xinerama.c#L313). To follow it farther than this I'll need to figure out X11's internal messaging, but my guess is that Xinerama's numbering is independent of the numbering that gdk uses via `ScreenOfDisplay`.

echarliewhite:
Avoiding use of PagerHints would require either XMonad support for additional EWMH hints (as discussed in https://github.com/xmonad/xmonad-contrib/issues/150) or getting the active monitor via `_NET_ACTIVE_WINDOW` (which wouldn't work for empty workspaces)

IvanMalison:
> Avoiding use of PagerHints would require either XMonad support for additional EWMH hints (as discussed in [xmonad/xmonad-contrib#150](https://github.com/xmonad/xmonad-contrib/issues/150)) or getting the active monitor via `_NET_ACTIVE_WINDOW` (which wouldn't work for empty workspaces)

or making the monitor/screen number stuff work properly, which would be the solution I would prefer.

IvanMalison:
I actually think that #399 is describing a behavior that is slightly different than what we have here btw.

The desired behavior there is to show as active on each monitor the workspace that is displayed on THAT monitor.

IvanMalison:
@echarliewhite any thoughts about finishing this?

colonelpanic8:
@rvl are you planning on trying to merge this?

rvl:
Not yet @colonelpanic8, and not if it requires changes to PagerHints. I was going to do some testing and see if I could work out the monitor number mapping.

rvl:
@echarliewhite, I followed those source code links you sent, thanks for the research. It seems to me unlikely that XMonad's screen numbering scheme will always match up with that of `gdk_display_get_monitor`. Usually it should, but probably not always, due to this: [`XMonad.Operations.getCleanedScreenInfo`](https://hackage.haskell.org/package/xmonad-0.18.0/docs/src/XMonad.Operations.html#nubScreens).

However, I don't think it's a problem for this feature.

XMonad tells us the active window id by EWMH, and then GDK can figure out which monitor it's on. If there is no active window (i.e. root window is active), then we can use the mouse pointer location ([example](https://github.com/taffybar/taffybar/blob/bf137ec6cdf3e853d4fa6d1699778f95ccce00ca/src/System/Taffybar/DBus/Toggle.hs#L58-L65)).

We know the active workspace by EWMH. If we know monitor number of the active window then we know the monitor number of the active workspace. The Taffybar on that monitor should highlight this workspace. Taffybars on other monitors shouldn't highlight this workspace.

Then we have the other visible workspace(s), known from `TaffybarPagerHints`.
 - If there are 0 other visible workspaces, then job done, nothing else to highlight.
 - If there is 1 other visible workspace then it's a dual-head setup. We can say the other monitor has that workspace, and the Taffybar on that monitor should highlight this workspace and not highlight any others.
 - If there are 2 or more other visible workspaces, then I'm not sure yet. Possibly the workspaces widget can remember the last active workspace that it highlighted.

PR #539, opened 2021-11-20
Add Active Icon next to window label in Windows Widget.

selene57:
Adds the option to show the active window icon next to the label for the windows widget.


src/System/Taffybar/Widget/Windows.hs
```diff
@@ -73,19 +83,37 @@ windowsNew :: WindowsConfig -> TaffyIO Gtk.Widget
 windowsNew config = do
   label <- lift $ Gtk.labelNew Nothing
 
+  icon <- lift $ Gtk.imageNew
+
+  context <- ask
```
IvanMalison, line 88: Lets get rid of the extra newlines here. looks kinda weird as is.

src/System/Taffybar/Widget/Windows.hs
```diff
@@ -62,6 +70,8 @@ defaultWindowsConfig =
   WindowsConfig
   { getMenuLabel = truncatedGetMenuLabel 35
   , getActiveLabel = truncatedGetActiveLabel 35
+  , showActiveIcon = True
+  , getActiveWindowIconPixbuf = defaultGetWindowIconPixbuf
```
IvanMalison, line 74: Instead of a Bool AND a getter, why not just a `Maybe WindowIconPixbufGetter`

IvanMalison: Also a better name would be getWindowIconPixbuf. This pixbuf getter could also be used in e.g. the menu.

src/System/Taffybar/Widget/Windows.hs
```diff
@@ -73,19 +83,37 @@ windowsNew :: WindowsConfig -> TaffyIO Gtk.Widget
...
   label <- lift $ Gtk.labelNew Nothing
 
+  icon <- lift $ Gtk.imageNew
+
+  context <- ask
+
   let setLabelTitle title = lift $ postGUIASync $ Gtk.labelSetMarkup label title
-      activeWindowUpdatedCallback _ = getActiveLabel config >>= setLabelTitle
+      getActiveWindowData = getActiveWindow >>= (traverse $ getWindowData Nothing [])
+      getThePixbuf size = fmap join $ (runX11Def Nothing getActiveWindowData) >>= (traverse $ getActiveWindowIconPixbuf config size)
```
IvanMalison, line 92: Call this `getActiveWindowPixbuf`

src/System/Taffybar/Widget/Windows.hs
```diff
@@ -73,19 +83,37 @@ windowsNew :: WindowsConfig -> TaffyIO Gtk.Widget
...
+      refreshImage = flip runReaderT context . getThePixbuf
```
IvanMalison, line 93: I don't think this part needs a name, just put this inline in the call to autoSizeImage below.

IvanMalison (commented):
A Changelog entry is needed.

IvanMalison:
@selene57 ping any thoughts on finishing this?

src/System/Taffybar/Widget/Windows.hs
rvl: Converted to `Maybe WindowIconPixbufGetter`.
We can't easily use the record name `getWindowIconPixbuf` because it conflicts with the one from `WorkspacesConfig`.

src/System/Taffybar/Widget/Windows.hs
rvl: Done.

src/System/Taffybar/Widget/Windows.hs
rvl: I moved the code into an auxiliary function `buildWindowsIcon` so that it looks neater.

rvl:
I have updated this with latest master and addressed review comments.

colonelpanic8 (approved):

PR #548, opened 2022-04-26
Updated stack.yaml to nightly-2021-04-26

jchia:
This gets GHC 9.2.2 together with aeson-2. The taffybar.cabal has been updated to require aeson >= 2.0.

I'm not sure how the nix stuff in the stack.yaml works so I left it alone.


IvanMalison:
@jchia can you check and see if something a bit more stock works now? I believe it should.

PR #553, opened 2022-09-05
Add CLI option `--project-name`/`-p` to customize taffybar configuration location

### LINKED ISSUE #552: Support running multiple/different bar configurations
Main proposal: add a CLI flag to taffybar's default launcher to customize which configuration to launch (instead of hardcoded `taffybar.hs`).

Here's my example use case. I want to run multiple bar instances with different definitions concurrently:
- the top bar contains a workspace switcher, window title
- the bottom bar contains widgets like brightness, volume, weather, CPU, FS monitor, etc., systray.
There's too many widgets to fit within a single bar (all on the top), so I like to have two separate bars.

(example view)
image

Coming from polybar, this is easy to setup-- just define two bar configurations, then pass in the name of the bar config as argument when launching polybar.  However, taffybar [hardcodes launching from `$XDG_CONFIG_HOME/taffybar/taffybar.hs`](https://github.com/taffybar/taffybar/blob/945a08452660de603193da8d297d559fdca497d1/app/Main.hs#L46), making it impossible to have different definitions and choose different ones to launch.

This is already sort of possible with taffybar, but it's not very convenient: I could just compile & launch the non-default taffybar instance manually from a different location (requires manually invoking `ghc` or `stack`, no support for change-detection and auto-recompile).  What I'm really looking for is native launching support (via dyre), with automatic change-detection and recompiling, and launching directly via the `taffybar` command instead of a separately-installed executable.

I propose adding some optional, backwards-compatible CLI flag to the main taffybar executable, such as `--config-main`, to be used as:
```bash
taffybar --config-main=taffybar-alt.hs
```
(where the current/default behavior when the argument is omitted is equivalent to passing `--config-main=taffybar.hs`), which customizes which taffybar configuration to launch via dyre.

---

To be clear, this is _different_ from https://github.com/taffybar/taffybar/issues/139, which (based on my understanding) seems to be about running multiple instances of the _same_ bar on multiple monitors.  My request is about choosing different bars to run at the same time, independent of multi-monitor setups.

kwshi:
Commits:

- gitignore: add dist-newstyle
- taffybar.cabal: expose dyre, xdg-basedir as direct dependencies
- app/Main.hs: add CLI option to customize dyre project name

Resolves #552.

Intended to used like so:

```bash
taffybar --project-name=PROJECTNAME
```

By default, when the option is omitted, the behavior is `--project-name=taffybar`. This launches the taffybar configuration defined at `$XDG_CONFIG_HOME/taffybar/taffybar.hs`, which is the existing behavior.

Meanwhile, if I wanted to define and run another bar instance, I could place it at `$XDG_CONFIG_HOME/taffybar-alt/taffybar-alt.hs`, and then run

```bash
taffybar --project-name=taffybar-alt
```

If I wanted to reuse library/source code across the two definitions, I can define the two bars in the same folder, then add a symlink `taffybar-alt`→`taffybar`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /dist/
+/dist-newstyle/
 .stack-work
 .cabal-sandbox
 cabal.sandbox.config
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -3,11 +3,13 @@
 -- recompile itself.
 module Main ( main ) where
 
+import qualified Config.Dyre as Dyre
 import Data.Default (def)
 import Data.Semigroup ((<>))
 import Data.Version
 import Options.Applicative
 import System.Directory
+import System.Environment.XDG.BaseDir
 import System.Log.Logger
 import System.Taffybar
 import System.Taffybar.Context
@@ -26,6 +28,16 @@ logP =
   <> value WARNING
   )
 
+projectNameP :: Parser String
+projectNameP =
+  option str
+  (  long "project-name"
+  <> short 'p'
+  <> help "Set taffybar configuration project name"
+  <> metavar "NAME"
+  <> value "taffybar"
+  )
+
 versionOption :: Parser (a -> a)
 versionOption = infoOption
                 (printf "taffybar %s" $ showVersion version)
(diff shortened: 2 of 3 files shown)


kwshi:
Per discussion in #552, this is apparently unnecessary, as it's already possible to do this two ways:
- Use an environment variable to launch different bar configurations as separate processes, or
- Use the advanced bar configuration (`barConfigGetter`) to run multiple bars in the same process.
Closing for now.

PR #555, opened 2022-10-09
workspaces: allow sorting windows by ewmh stack index

teleshoes:
xmonad-contrib 0.17.0 broke my icon sort order, which used focus history.

prior to xmonad-contrib 0.17.0, using `iconSort=return` sorted icons in focus history order, due to wrong impl of _NET_CLIENT_LIST. since 0.17.0, focus history is now (more correctly) in _NET_CLIENT_LIST_STACKING.
https://github.com/xmonad/xmonad-contrib/commit/fa82db1130258d83014ab6769a5297552c53242b

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@
 
  * Window titles are shown on hover.
 
+ * Allow sorting workspace window icons by _NET_CLIENT_LIST_STACKING.
+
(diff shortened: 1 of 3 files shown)


IvanMalison:
Looks great. Can you add a CHANGELOG entry?

PR #576, opened 2024-05-01
Build with ghc96

colonelpanic8:
@jchia This works for me.

Are you building with stack or raw cabal? cabal sandboxes?

### DIFF
--- a/comp.nix
+++ b/comp.nix
@@ -1,1 +1,1 @@
-{ compiler = "ghc94"; }
+{ compiler = "ghc96"; }
--- a/flake.lock
+++ b/flake.lock
@@ -21,11 +21,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1689068808,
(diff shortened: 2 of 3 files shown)


jchia:
I use stack.

GHC 9.6.5 ships with mtl-2.3:
https://downloads.haskell.org/~ghc/9.6.5/docs/users_guide/9.6.5-notes.html (See 'Included libraries')

gi-cairo-connector-0.1.1 rejects mtl-2.3:
https://hackage.haskell.org/package/gi-cairo-connector-0.1.1

I don't know how the nix build works. Perhaps it disregards dependency version bounds?

colonelpanic8:
@jchia ah my bad, here is a new revision: https://hackage.haskell.org/package/gi-cairo-connector-0.1.1/revision/1.cabal

jchia:
It works, but there is a new problem. The latest version of scotty, scotty-0.22, is rejected by taffybar. I'll file another issue for that.

PR #583, opened 2024-09-19
flake.nix: Build for multiple compiler versions, add checks

rvl:
1. Adds flake outputs building taffybar against defined ghc versions from nixpkgs.
2. Adds some flake checks for HLint, `-Werror`.
3. Splits the nixpkgs overlay into a separate `overlay.nix` which can be consumed by non-flake projects.
4. Puts back the cachix settings in `flake.nix`.
5. Update the github actions workflow to build more stuff.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,2 @@
+watch_file taffybar.cabal nix/*.nix
 use flake .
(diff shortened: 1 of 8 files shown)


rvl:
The github actions fail at HLint, which is expected.

colonelpanic8:
@rvl can you elaborate about why?

rvl:
Just because there were 20 HLint suggestions, but they were only fixed in #584.

I have rebased this branch now, so it should pass.

PR #601, opened 2024-11-05
Reload CSS files on SIGHUP

rvl:
This adds a way to reload CSS without restarting Taffybar.

Restarting Taffybar is quick, but it does make windows flicker a bit, which is not so good for observing results of style changes.

Perhaps in future a D-Bus method could be added, or the files could be watched by hfsnotify. But for me a signal is enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,12 @@
 
  * Add icon next to window label in Windows widget. This can be configured with [`WindowsConfig(getActiveWindowIconPixbuf)`][WindowsConfig].
 
- * If Taffybar is not running in a terminal, then it will reload CSS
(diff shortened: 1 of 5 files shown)


colonelpanic8:
why not monitor the file with inotify and automatically update if the file changes?

rvl:
> why not monitor the file with inotify and automatically update if the file changes?

Yeah that would be the best way and it would be good to implement in the future.
I just want to do the first step - manual reloading - for now.
There are a few details to take care of with inotify.

rvl:
OK @colonelpanic8, I have added automatic reloading with inotify.

colonelpanic8 (approved):

colonelpanic8 (approved):

PR #603, opened 2024-11-10
Fix a few lints

colonelpanic8:


rvl (approved):

rvl:
Is `StrictData` used so that X11 properties are eagerly fetched? Or another reason?

Issue #600, opened 2024-11-04
GHC 9.8


jchia:
**Describe the bug**
The project is described in `taffybar.cabal` as having been tested with GHC 9.8.2. But apparently it does not work with GHC 9.8.2.

**To Reproduce**
`cabal build` fails because broadcast-chan does not support GHC 9.8.2:
```
$ cabal build
Resolving dependencies...
Error: [Cabal-7107]
Could not resolve dependencies:
[__0] trying: my-taffybar-0.1.0.0 (user goal)
[__1] trying: base-4.19.1.0/installed-c38a (dependency of my-taffybar)
[__2] trying: taffybar-4.0.3 (user goal)
[__3] next goal: broadcast-chan (dependency of taffybar)
[__3] rejecting: broadcast-chan-0.2.1.2 (conflict: base==4.19.1.0/installed-c38a, broadcast-chan => base>=4.8 && <4.19)
[__3] skipping: broadcast-chan; 0.2.1.1, 0.2.1, 0.2.0.2, 0.2.0.1, 0.2.0 (has the same characteristics that caused the previous version to fail: excludes 'base' version 4.19.1.0)
[__3] rejecting: broadcast-chan-0.1.1 (conflict: taffybar => broadcast-chan>=0.2.0.2)
[__3] skipping: broadcast-chan-0.1.0 (has the same characteristics that caused the previous version to fail: excluded by constraint '>=0.2.0.2' from 'taffybar')
[__3] fail (backjumping, conflict set: base, broadcast-chan, taffybar)
After searching the rest of the dependency tree exhaustively, these were the goals I've had most trouble fulfilling: base, taffybar, broadcast-chan, my-taffybar
Try running with --minimize-conflict-set to improve the error message.
```
However, even after commenting out broadcast-chan from `taffybar.cabal`, `cabal build` inexplicably fails on `tls`:
```
[39 of 76] Compiling Network.TLS.Parameters ( Network/TLS/Parameters.hs, dist/build/Network/TLS/Parameters.o, dist/build/Network/TLS/Parameters.dyn_o )

Network/TLS/Parameters.hs:424:35: error: [GHC-39999]
    • No instance for ‘Default ValidationCache’
        arising from a use of ‘def’
    • In the ‘sharedValidationCache’ field of a record
      In the expression:
        Shared
          {sharedCredentials = mempty,
           sharedSessionManager = noSessionManager, sharedCAStore = mempty,
[...]

PR #604, opened 2024-12-19
Replaced BroadcastChan With TChan

jchia:
This partially addresses #600 by removing the dependency on broadcast-chan, which seems to be practically unmaintained.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,13 @@
    a `SIGHUP` signal, then it will restart the inotify instance and
    reload the CSS files.
 
+ * The [`BroadcastChan`][broadcast-chan] dependency has not received
+   Cabal revisions for a while, so is replaced with [`Control.Concurrent.STM.TChan`][tchan].
+   Some types in `System.Taffybar.Hooks` and `System.Taffybar.Information.{Battery,Chrome,Crypto}` change accordingly.
+
 [WindowsConfig]: https://hackage.haskell.org/package/taffybar-4.0.4/docs/System-Taffybar-Widget-Windows.html#t:WindowsConfig
+[broadcast-chan]: https://hackage.haskell.org/package/broadcast-chan
+[tchan]: https://hackage.haskell.org/package/stm-2.5.1.0/docs/Control-Concurrent-STM-TChan.html
 
 
 # 4.0.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 4.0.4 (UNRELEASED)
+# 4.1.0 (UNRELEASED)
 
 ## Improvements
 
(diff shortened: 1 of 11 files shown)


rvl:
Thanks @jchia , I'll have a look :eyes:.

colonelpanic8:
Thanks @jchia I don't know about the exact semantics of TChan. One thing I would be curious about is what it does in the case where there are no readers. IIRC the reason that I elected to use broadcast chan was because it avoided space leaks in this case.

colonelpanic8:
I'm definitely open to this, but I think we need to test it extensively to ensure that everything behaves in the same way as it did before.

I'd also be open to forking broadcast chan in order to make it compatible with newer versions of ghc.

jchia:
> Thanks @jchia I don't know about the exact semantics of TChan. One thing I would be curious about is what it does in the case where there are no readers. IIRC the reason that I elected to use broadcast chan was because it avoided space leaks in this case.

Where did you hear about the leaks? Was it just from the broadcast-chan top-level documentation? If so, the documentation was describing `Control.Concurrent.Chan`, not `Control.Concurrent.STM.TChan`. IMO, the only good reason for a channel to leak space is that it is accumulating/buffering items for a future reader to read. AFAICT, `Control.Concurrent.STM.TChan` does not do this.

Here's a test program to elucidate the behavior: https://github.com/jchia/tchan

Test 1 shows you that readers do not start reading from the very beginning. You can only write to a channel created with `newBroadcastTChan`. A `reader` channel needs to be created from the `writer` channel by calling `dupTChan`. A reader reads only items that are written after the reader channel is created, not items that were written before the reader channel was created.

Test 2 shows you that the memory usage is limited; while you run Test 2, you can run `top` and see that the resident memory is not increasing nearly as much as would be necessary if all items written while there are no readers are being buffered and causing a space leak. You could also change the code so that the readers are never created to be sure.

jchia:
Any updates? I've been regularly using my version for a few weeks, though I don't know which aspects of Taffybar actually use broadcast channels. IMO, a fork of broadcast-chan would entail quite a lot of misplaced development maintenance effort and headache, whereas TChan is in stm, which comes with GHC. In other words, it would be reinventing the wheel.

jchia:
One improvement that can be made is to have type-safe channels, i.e. distinct types for reading and writing respectively, like in broadcast-chan and unlike in TChan. I can implement that if it's decided that taffybar is to use TChan instead of broadcast-chan.

rvl (approved):
Tested this branch and it works, thanks @jchia.

I like that `BroadcastChan` is closeable, but that's not enough reason to keep it if the dependency causes friction for users and maintainers.

I added a ChangeLog entry.

jchia:
Would you like the newtype wrapper I was talking about?

PR #608, opened 2025-07-18
Add support of refreshing in PollingLabel (mostly for WttrIn)

Commelina:
### Motivation
The WttrIn widget can sometimes encounter network errors or display outdated information. It will be convenient to add a "refreshing" function on these scenes rather than restarting the whole bar, as shown in the figures.

- Network error: image
- Click to refresh the widget: image
- Reloaded result: image

### Changes
1. The PR adds a 
   ```
   pollingLabelWithVariableDelayAndRefresh
     :: IO (T.Text, Maybe T.Text, Double)
     -> Bool -- ^ Whether to refresh the label on mouse click
     -> m GI.Gtk.Widget
   ```
   function in `System.Taffybar.Widget.Generic.PollingLabel` module. The previous `pollingLabelWithVariableDelay` function remains the same and is a special version of `pollingLabelWithVariableDelayAndRefresh`. 
2. The WttrIn widget created by `textWttrNew` of `System.Taffybar.Widget.WttrIn` can be manually refreshed by a single click.

### DIFF
--- a/src/System/Taffybar/Widget/Generic/PollingLabel.hs
+++ b/src/System/Taffybar/Widget/Generic/PollingLabel.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE OverloadedStrings #-}
+
 -- | This is a simple text widget that updates its contents by calling
 -- a callback at a set interval.
 module System.Taffybar.Widget.Generic.PollingLabel where
@@ -8,6 +10,7 @@ import           Control.Monad
 import           Control.Monad.IO.Class
 import qualified Data.Text as T
 import           GI.Gtk
+import qualified GI.Gdk as Gdk
 import           System.Log.Logger
 import           System.Taffybar.Util
 import           System.Taffybar.Widget.Util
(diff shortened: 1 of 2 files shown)


src/System/Taffybar/Widget/Generic/PollingLabel.hs
```diff
@@ -48,9 +51,27 @@ pollingLabelWithVariableDelay
   => IO (T.Text, Maybe T.Text, Double)
   -> m GI.Gtk.Widget
 pollingLabelWithVariableDelay action =
+  pollingLabelWithVariableDelayAndRefresh action False
+
+-- TODO: Customize the delay and message on mouse click
+pollingLabelWithVariableDelayAndRefresh
```
colonelpanic8, line 57: Why not wrap this in the other direction?

colonelpanic8: i.e. make it so that the pollingLabelWithVariableDelayAndRefresh wraps  pollingLabelWithVariableDela

Commelina: Uh, both of them are possible options. My choice is because some other modules in this project are also in this style (e.g. `pollingGraphNew` wraps `pollingGraphNewWithTooltip` in `System.Taffybar.Widget.Generic.PollingGraph`). And it is common to write a most general version of a function and wrap others with it in many Haskell packages like GHC `base`. What's more, this can merge the "internal" details into one place rather than scattering them in many functions.

